PR #18, opened 2017-12-18
Support excludeFromIndexes for primitives in arrays.

### LINKED ISSUE #13: Datastore: field types of array are always showing indexed regardless of excludeFromIndexes
###### Copied from original issue: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2615

<a href="/DaveyEdwards"></a><a href="/DaveyEdwards">@&shy;DaveyEdwards</a><br>September 16, 2017 8:53 PM

Saving array's shows it is indexed (not all the actual items in the array, just the field), even when excludeFromIndexes: true.

I am not sure if this is normal behavior and the whole array is not actually stored in the index.  I had asked this originally in gstore-node issues and he said to ask here :)

## You can see the "circles" field shows indexed
image

## Inside the "circles" it is showing all array items are excluded from index
image

#### Environment details

  - OS: Windows 10
  - Node.js version: 6.11.2
  - npm version: 3.10.10
  - google-cloud-node version: 
 Replecated on google-cloud: 0.56.0
and @google-cloud/datastore: 1.1.0
I originally was using gstore-node and saw this problem, and now I am using just raw datastore and still have it.

#### Steps to reproduce
Seed database with a field of array

## My test function
```js
createEntity([
  {
    name: 'kind',
    value: 'Pages',
    excludeFromIndexes: true,
  },
  {
    name: '_id',
    value: '6',
  },
  {
    name: 'title',
    value: 'my first title',
  },
  {
    name: 'dateCreated',
    value: new Date(),
  },
  {
    name: 'string',
    value: 'A long string goes here',
    excludeFromIndexes: true,
  },
  {
    name: 'circles',
    value: ['1', '2', '3', '4'],
    excludeFromIndexes: true,
  },
]);
```
The actual save function
```js
export default async function createEntity(entity) {
  let response = null;
  let kind = null;
  let dsKey = null;

  try {
    entity.map((entityFeilds) => {
      if (entityFeilds.name === '_id') {
        dsKey = entityFeilds.value;
        if (dsKey === '') {
          dsKey = null;
        }
        return dsKey;
      }
      if (entityFeilds.name === 'kind') {
        kind = entityFeilds.value;
        return kind;
      }
      return null;
    });
    const key = datastoreClient.key([kind, dsKey]);
    const data = entity;
    const newEntity = {
      key,
      data,
    };

    response = await datastoreClient.save(newEntity);
  } catch (err) {
    console.error('save err', err);
  }
  return response;
}
```

beaulac:
Fixes #13. PR ported from [this one](https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2773). In addition, adds an example to the documentation.

- [x] Tests and linter pass

### DIFF
--- a/src/entity.js
+++ b/src/entity.js
@@ -499,9 +499,14 @@ function entityToEntityProto(entityObject) {
     var hasEntityPath = entityIndex > -1;
 
     if (!hasArrayPath && !hasEntityPath) {
-      if (entity.properties[path]) {
-        // This is the property to exclude!
-        entity.properties[path].excludeFromIndexes = true;
+      if (entity.properties) {
+        if (entity.properties[path]) {
+          // This is the property to exclude!
+          entity.properties[path].excludeFromIndexes = true;
+        }
+      } else if (!path) {
+        // This is a primitive that should be excluded.
+        entity.excludeFromIndexes = true;
       }
       return;
     }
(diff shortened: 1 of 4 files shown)


stephenplusplus:
Thank you 🙏 Would you mind also copying over the change to the system-test file? https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2773/files#diff-def9264f47890d5d0e327fb24796179b

beaulac:
Just did it; had forgotten about that part.

PR #28, opened 2017-12-21
Provide user's project ID to google-auto-auth.

### LINKED ISSUE #27: Cannot connect to the emulator without default credentials file
#### Environment details

  - OS: OSX 10.12.6
  - Node.js version: v8.9.3
  - npm version: 5.5.1
  - @google-cloud/datastore version: 1.3.2

Updated debug info [here](https://github.com/Sobesednik/datastore-emulator-test#google-clouddatastore132-without-default-credentials).

#### Steps to reproduce

  1. `git clone https://github.com/Sobesednik/datastore-emulator-test.git`
  1. `cd datastore-emulator-test`
  1. `npm i`
  1. `gcloud beta emulators datastore start --no-store-on-disk`
  1. (new termial) `$(gcloud beta emulators datastore env-init)`
  1. `unset GOOGLE_APPLICATION_CREDENTIALS`  
  1. `npm t`

This used to work in `@google-cloud/datastore@1.1.0`. If I just want to test locally on the emulator, do I have to have a credentials file now by design?

stephenplusplus:
Fixes #27 

When we switched to GAPIC, we put a funnel method in front of all API requests before they were sent to GAPIC. From that, we get the `projectId` from google-auto-auth, to fill in any placeholders. We overlooked a scenario where a projectId is provided by the user already, such as when they are using a local emulator.

After this PR, we will honor that project ID, and not ask google-auto-auth to try to grab it from a credentials file (which is not generally provided when users are working from a local emulator).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -370,17 +370,22 @@ function Datastore(options) {
 
   this.clients_ = new Map();
   this.datastore = this;
+
   /**
    * @name Datastore#namespace
    * @type {string}
    */
   this.namespace = options.namespace;
+
+  const userProvidedProjectId =
+    process.env.DATASTORE_PROJECT_ID || options.projectId;
+  const defaultProjectId = '{{projectId}}';
+
   /**
    * @name Datastore#projectId
    * @type {string}
    */
-  this.projectId =
-    process.env.DATASTORE_PROJECT_ID || options.projectId || '{{projectId}}';
+  this.projectId = userProvidedProjectId || defaultProjectId;
 
   this.defaultBaseUrl_ = 'datastore.googleapis.com';
   this.determineBaseUrl_(options.apiEndpoint);
@@ -392,6 +397,7 @@ function Datastore(options) {
       scopes: gapic.v1.DatastoreClient.scopes,
       servicePath: this.baseUrl_,
       port: is.number(this.port_) ? this.port_ : 443,
+      projectId: userProvidedProjectId,
     },
     options
   );
--- a/src/index.js
+++ b/src/index.js
@@ -378,7 +378,7 @@ function Datastore(options) {
   this.namespace = options.namespace;
 
   const userProvidedProjectId =
-    process.env.DATASTORE_PROJECT_ID || options.projectId;
+    options.projectId || process.env.DATASTORE_PROJECT_ID;
   const defaultProjectId = '{{projectId}}';
 
   /**
@@ -401,7 +401,6 @@ function Datastore(options) {
     },
     options
   );
-
   if (this.customEndpoint_) {
     this.options.sslCreds = grpc.credentials.createInsecure();
   }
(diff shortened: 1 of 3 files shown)


test/index.js
```diff
@@ -222,6 +227,7 @@ describe('Datastore', function() {
             scopes: v1.DatastoreClient.scopes,
             servicePath: datastore.baseUrl_,
             port: 443,
+            projectId: undefined,
```
alexander-fenster: This test will fail if the environment has DATASTORE_PROJECT_ID defined for any reason (e.g. in the user profile). Can we explicitly delete it for this test? Also, I would add a test that would verify that the value from DATASTORE_PROJECT_ID comes to datastore.options (if there is no such test).

stephenplusplus: Good call, done.

src/index.js
```diff
@@ -370,17 +370,22 @@ function Datastore(options) {
...
   this.datastore = this;
+
   /**
    * @name Datastore#namespace
    * @type {string}
    */
   this.namespace = options.namespace;
+
+  const userProvidedProjectId =
+    options.projectId || process.env.DATASTORE_PROJECT_ID;
```
stephenplusplus: This is flipped from how it was before, which I think is what would be expected. If the user manually provides a projectId, I think we're supposed to honor that and not do any magic. What do you think?

test/index.js
```diff
@@ -222,12 +239,15 @@ describe('Datastore', function() {
...
           },
           options
         )
       );
     });
 
+    it('should cache the correct projectId', function() {});
```
alexander-fenster: why do we need this test?

stephenplusplus: Woops.

src/index.js
alexander-fenster: My understanding is that environment variable should always be prioritized over anything written in the code. In my world, command line option > environment variable > local configuration file > global configuration file > default value in the source code. You might have different opinion though :)

stephenplusplus: I went back to an [old issue]((https://github.com/GoogleCloudPlatform/google-cloud-node/issues/191) where we discussed a similar topic, and it reminded me of a feature we offer in our libraries... the user can uniquely authenticate multiple instantiations. So, I can hold an instance of `Datastore` for "project-a" and another for "project-b". We let the user choose at run-time which project each invocation refers to.

> command line option > environment variable > local configuration file > global configuration file > default value in the source code

I would agree with that, however, I believe you've ranked `options.projectId` as the "local configuration file". To me, the user explicitly writing "projectId: 'my-project'" is the same rank as a command line option. 

I think that if the user typed out "my-project", but we ignore that and use an env var, that would be unexpected for them.

alexander-fenster: OK, let's leave it this way (probably until someone complains :) )

PR #48, opened 2018-01-31
Don't modify Key constructor input path.

### LINKED ISSUE #47: Key constructor mangles input path
The `entity.Key` constructor modifies the `path` that is used to build a `Key`. 

This causes problems if references to the path are used later, and is not generally desired/expected constructor behaviour. Also, an errors occurs if the `path` is sealed/frozen.

#### Steps to reproduce

```javascript
const ds = require('@google-cloud/datastore')(/* ... */);

const parentPath = ['parentKind', 'identifier'];
const parentKey = ds.key(keyPath);

// parentPath was modified by constructor side-effects:
const parentKind = parentPath[0]; // undefined;

const childPath = [...parentPath, 'kind', 'identifier'];
const childKey = ds.key(childPath);

// The problem:
const expected = (childKey.path[0] === 'parentKind');
const actual = (childKey.path[0] === 'kind');

// Also:
ds.key(Object.seal(['kind', 'identifier'])) // TypeError: Cannot delete property ...
```

#### Environment details

  - OS: macOS 10.13.3
  - Node.js version: 8.9.1
  - npm version: 5.6.0
  - google-cloud/datastore version: 1.3.3

beaulac:
Fixes #47. Also corrects some inconsistencies in Key constructor JSDocs. 

- [x] Tests and linter pass
- [x] Code coverage does not decrease
- [x] Appropriate docs were updated

### DIFF
--- a/src/entity.js
+++ b/src/entity.js
@@ -158,9 +158,9 @@ entity.isDsGeoPoint = isDsGeoPoint;
  * Build a Datastore Key object.
  *
  * @class
- * @param {object} options - Configuration object.
- * @param {array} options.path - Key path.
- * @param {string} [options.namespace] - Optional namespace.
+ * @param {object} options Configuration object.
+ * @param {array} options.path Key path.
+ * @param {string} [options.namespace] Optional namespace.
  *
  * @example
  * const Datastore = require('@google-cloud/datastore');
--- a/src/entity.js
+++ b/src/entity.js
@@ -158,9 +158,9 @@ entity.isDsGeoPoint = isDsGeoPoint;
  * Build a Datastore Key object.
  *
  * @class
- * @param {object|string|array} Configuration object.
- * @param {string|array} options.path  Key path.
- * @param {string} [options.namespace] Optional namespace.
+ * @param {object} options - Configuration object.
+ * @param {array} options.path - Key path.
+ * @param {string} [options.namespace] - Optional namespace.
  *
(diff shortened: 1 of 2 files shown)


src/entity.js
```diff
@@ -178,7 +178,7 @@ function Key(options) {
   this.namespace = options.namespace;
```
stephenplusplus: I think we could just add this line around here:

```js
options.path = [].slice.call(options.path);
```

stephenplusplus:
Thanks for the PR! I left a thought for a simplification. Let me know what you think.

src/entity.js
beaulac: That is indeed a lot better; I'll make this change 👍

src/entity.js
```diff
@@ -158,8 +158,8 @@ entity.isDsGeoPoint = isDsGeoPoint;
  * Build a Datastore Key object.
  *
  * @class
- * @param {object|string|array} Configuration object.
- * @param {string|array} options.path  Key path.
+ * @param {object} options Configuration object.
+ * @param {array} options.path Key path.
```
stephenplusplus: Thanks for catching these mis-types. It looks like the rest of the docs markup doesn't use hyphens, so I took them out.

beaulac: Great! I'll keep that in mind for any future doc changes 👍

PR #71, opened 2018-03-19
v1.4.0

callmehiphop:


alexander-fenster:
The version of datastore in `samples/package.json` should also be updated. Also, not sure if you did it for other packages. Samples should be linked to the current version, otherwise, samples tests will test old version from npm and will not trigger errors if anything is wrong with the current package.

PR #101, opened 2018-06-08
excludeFromIndexes: fix for entity elements inside arrays

### LINKED ISSUE #93: Datastore save() fails on null value when excludeFromIndexes is used for nested arrays
#### Environment details

  - OS: Debian (rodete)
  - Node.js version: 8.9.4
  - npm version: 5.6.0
  - `@google-cloud/datastore` version: 1.4.0

#### Steps to reproduce

Execute the following code (snippet taken from a tiny express app to demo the issue):

```javascript
router.get('/save', function(req, res, next)  {
  const datastoreClient = Datastore();
  const entity = {
    key: datastoreClient.key(['test']),
    data: {
      "rules": [
        {"requirements": [{"audiences": "value"}]},
        {"requirements": null},
      ]
    },
    excludeFromIndexes: [
      'rules[].requirements[].audiences'
    ]
  };
  datastoreClient.save(entity)
      .then(() => { res.send('OK!'); })
      .catch(error => { res.send('Failed: ' + error); });
});
```

Result:
`Failed: TypeError: Cannot read property 'values' of undefined`

With some digging, it seems it's failing here:
https://github.com/googleapis/nodejs-datastore/blob/c1cfe5bfce2ba018c3d2b4e5d03fd1c6e2d420b8/src/entity.js#L539

This only happens with the combination of this data structure + this `excludeFromIndexes` value, since it causes a traversal of the object, including an expectation that the second `requirements` field is also an array.

It's sort of a weird edge case, but one we encountered in production this week.

### LINKED ISSUE #100: `excludeFromIndexes: ['path[].to.some.array[]']` should exclude (the roots of) entity elements belonging to the array
- `@google-cloud/datastore` version: 1.4.0, or [master@445304c7a9b97...](https://github.com/googleapis/nodejs-datastore/commit/445304c7a9b97b555e2f0dcdf5ccfd63e5ac290d) (same same...)

#### Steps to reproduce

  1. 
```
  entity = {
    key: key,
    data: {
      array: [
        element1: 'string',
        element2: { some: 'object' }
        element3: 1337,
      ]
    },
    excludeFromIndexes: ['array[]']
  };
  datastore.save(entity);
```
  2. The property 'array' of saved entity above shows up as 'indexed' in Datastore Console. The reason seems to be as follows:
  An array property apparently counts as excluded from indexes if and only if every element is excluded from indexes. (Other than entity and primitive properties, an array property can not be marked as excluded from indexes by itself, [ because semantically, it is supposed to just be some kind of "soft" container for one single "multi-valued" property, I guess ]. )
  The excludeFromIndexes logic in `entityToEntityProto` (source file [`src/entity.js`](https://github.com/googleapis/nodejs-datastore/blob/master/src/entity.js), latest update from PR [#2773 (old repo)](https://github.com/GoogleCloudPlatform/google-cloud-node/pull/2773) merged in #18) currently excludes *only primitive elements* of an array that is marked with `"excludeFromIndexes: ['path.to.array[]']"`. Entity elements included in the array are simply skipped, and not excluded from indexing. Therefore, the array itself as a unit is not excluded from indexing, if it contains entity elements, despite marking it with `"excludeFromIndexes: ['path.to.array[]']"`.
  This seems illogical to me. It is perfectly possible to exclude entity elements inside array properties from indexing (which does not imply excluding all the nested properties of said entity elements from indexing, but only the root). It can be done with the low-level API for entity elements included in arrays. Those array properties, if all their individual elements are excluded, will then show up as 'not indexed' in the Datastore Console, as it should be, and will not waste index space.

  I will submit a simple PR for this.

**PS:** Currently there are two different supported syntaxes to exclude entity properties from indexing.
  a) "Array syntax" (apparently intended to be deprecated at some point):
``` entity = {
  key: key,
  data: [{
    name: 'rootProperty1',
    value: 'some property at the root',
    excludeFromIndexes: true/false
  }, {
    name: 'rootProperty2',
    value: [ 'why', 'not', 'an', 'array'],
    excludeFromIndexes: false
  }, {
    name: 'rootProperty3',
    value: {
      nestedProperty: 'nasty'
    },
    excludeFromIndexes: true
  }
}]
};
```
  This syntax only allows to specify "excludeFromIndexes" as true or false for each single property at the root. The current behaviour of this syntax for arrays is: If an array property is marked with `excludedFromIndexes: true`, then all its elements are marked as excluded from indexes, be they primitives or entities (the corresponding entity elements' sub-properties are not automatically excluded from indexes thereby). This is consistent with the behaviour of my pull-request for the other syntax.

  I don't know if it is possible in some convoluted way with this syntax to mark sub- and sub-sub-properties and -arrays as excluded from indexes as well. I believe it is not. It seems this "old" syntax allows only for excluding first-level children/properties (including arrays) from indexing.

  b) The "new" (and apparently favoured for the future) syntax referred to [in this PR #101 here](https://github.com/googleapis/nodejs-datastore/pull/101):
``` entity = {
  key: key,
  data: {
    just: { a: 'simple object', with: ['all', 'possible', 'kinds', 'of', 'nested', 'properties']}
}
  excludedFromIndexes: ['paths[].to', 'every', 'single.property[].that', 'should.be[].excluded', 'from.indexing[]'];
};
```

This syntax would allow, on principle, the exclusion from indexes of properties at any level... were it just not for this bug: With the current buggy behaviour, it is not possible to exclude the roots of arrays containing entity elements from indexing. It is only possible to exclude primitive array elements from indexing, and to exclude all sub-properties of entity array elements from indexing, but not to exclude the entity array elements (and therefore the containing array property) themselves.

trollkotze:
Fixes #93
Fixes #100

- [x] Tests and linter pass

### DIFF
--- a/test/entity.js
+++ b/test/entity.js
@@ -603,6 +603,9 @@ describe('entity', function() {
       var value10 = 'Stephen10';
       var value11 = 'Stephen11';
       var value12 = 'Stephen12';
+      var value13 = 'something';
+      var value14 = 'else';
+      var value15 = 'whatever';
 
       var entityObject = {
         excludeFromIndexes: [
@@ -615,7 +618,8 @@ describe('entity', function() {
           'array[].array[].entity.name',
           'entityExcluded[].name',
           'primitiveExcluded[]',
-          'rules[].requirements[].audiences'
+          'rules[].requirements[].audiences',
+          'nestedArrayVariants[].a[].b'
         ],
 
         data: {
@@ -676,7 +680,17 @@ describe('entity', function() {
             {
               requirements: null
             },
-          ]
+          ],
+
+          nestedArrayVariants: [
+            {
+              a: [{b:value13}, {c:value14}],
+            },{
+              a: null,
+            },{
+              a: [value15],
+            },
+          ],
 
         },
       };
@@ -847,6 +861,62 @@ describe('entity', function() {
                     properties: {
                       requirements: {
                         nullValue: 0
+                      },
+                    },
+                  },
+                },
+              ],
+            },
+          },
+          nestedArrayVariants: {
+            arrayValue: {
+              values: [
+                {
+                  entityValue: {
+                    properties: {
+                      a: {
+                        arrayValue: {
+                          values: [
+                            {
+                              entityValue: {
+                                properties: {
+                                  b: {
+                                    stringValue: value13,
+                                    excludeFromIndexes: true,
+                                  },
+                                },
+                              },
+                            },{
+                              entityValue: {
+                                properties: {
+                                  c: {
+                                    stringValue: value14,
+                                  },
+                                },
+                              },
+                            },
+                          ],
+                        },
+                      },
+                    },
+                  },
+                },{
+                  entityValue: {
+                    properties: {
+                      a: {
+                        nullValue: 0,
+                      },
+                    },
+                  },
+                },{
+                  entityValue: {
+                    properties: {
+                      a: {
+                        arrayValue: {
+                          values: [{
+                            stringValue: value15,
+                          }],
+                        },
                       },
                     },
                   },
--- a/test/entity.js
+++ b/test/entity.js
@@ -619,7 +619,7 @@ describe('entity', function() {
           'entityExcluded[].name',
           'primitiveExcluded[]',
           'rules[].requirements[].audiences',
-          'nestedArrayVariants[].a[].b'
+          'nestedArrayVariants[].a[].b',
         ],
 
         data: {
@@ -673,25 +673,26 @@ describe('entity', function() {
             {
               requirements: [
                 {
-                  audiences: value12
-                }
-              ]
+                  audiences: value12,
+                },
+              ],
             },
             {
-              requirements: null
+              requirements: null,
             },
           ],
 
           nestedArrayVariants: [
             {
-              a: [{b:value13}, {c:value14}],
-            },{
+              a: [{b: value13}, {c: value14}],
+            },
+            {
               a: null,
-            },{
+            },
+            {
               a: [value15],
             },
           ],
-
         },
       };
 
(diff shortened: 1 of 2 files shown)


test/entity.js
```diff
@@ -591,7 +591,17 @@ describe('entity', function() {
     });
 
     it('should respect excludeFromIndexes', function() {
-      var value = 'Stephen';
+      var value1 = 'Stephen';
```
alexander-fenster: I love this :)

alexander-fenster:
@trollkotze Welcome, and thanks for your contribution!

@stephenplusplus that's your code I guess, can you please take a look?

test/entity.js
trollkotze: That's the Stephen++ clone army. It keeps on incrementing. :D 
Just found it easier for testing with different values in different places.

trollkotze:
I have now also added a [simple fix](https://github.com/googleapis/nodejs-datastore/pull/101/commits/6d02872e8a9e43e86ec41f3d08133a4baed6871c) for issue #93, with the example provided in that issue as a test.
This allows now something like:
```
entity = {
  key: key,
  excludeFromIndexes: ['arr[].a[].b'],
  data: {
    nestedArrayVariants: [
      {
        a: [{b:'something'},{c:'else'}],
      },{
        a: null,
      },{
        a: ['whatever'],
      }
    ],
  },
};
```

Previously, `entityToEntityProto` would have thrown a type error when encountering `arr[i].a == null` (or any other primitive value) while trying to exclude `arr[].a[].b` from indexes, thus disallowing detailed exclusion below nested properties that might sometimes be arrays and sometimes something else.
The fix is one simple property existence check that will rarely be traversed.

stephenplusplus (approved):

stephenplusplus:
Awesome, thanks @trollkotze!

evgenius:
Hi guys, is there any chance this fix is released soon?
I faced the same issue and even started working on a patch but then realised it's already here but just not released yet.

PR #112, opened 2018-06-22
refactor: update nodejs-common and google-auth-library

JustinBeckwith:


src/index.js
```diff
@@ -404,7 +404,7 @@ function Datastore(options) {
     this.options.sslCreds = grpc.credentials.createInsecure();
   }
 
-  this.auth = googleAuth(this.options);
+  this.auth = new GoogleAuth(this.options);
```
stephenplusplus: There is a use of this in `src/request.js`, that I don't think is compatible with `GoogleAuth`: https://github.com/googleapis/nodejs-datastore/blob/8ef7e74a540a87b49a9bc6ce46d7947e071403e9/src/request.js#L1133

JustinBeckwith: I think it's ok.  `getDefaultProjectId` will accept a callback:
https://github.com/google/google-auth-library-nodejs/blob/master/src/auth/googleauth.ts#L137

And it's aliased to `getProjectId`:
https://github.com/google/google-auth-library-nodejs/blob/master/src/auth/googleauth.ts#L91

JustinBeckwith: Forgot to add - all the system tests pass :)

stephenplusplus: Nice! I tested in the console, but must have done something wrong.

PR #123, opened 2018-07-02
properly returns integer values for integer keys vs the string value

richardowright:
When an integer valued key (i.e. with id) was returned, the user would get a string vs. an integer. This creates inconsistencies and a need for significant coercion as using that value in a query would result in the wrong type (i.e. a string vs. the saved int).

- [x] Tests and linter pass

### DIFF
--- a/src/entity.js
+++ b/src/entity.js
@@ -657,7 +657,10 @@ function keyFromKeyProto(keyProto) {
 
     // only convert the id to an Integer Type object if it's not already a number
     if (path.idType === 'id') {
-      id = parseInt(id, 10) || new entity.Int(id);
+      var parsedInteger = parseInt(id, 10);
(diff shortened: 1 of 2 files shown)


stephenplusplus:
It looks like this would introduce more inconsistency, since the user wouldn't be sure when to handle Datastore.Int's vs a Number. I don't prefer returning the custom type either, however, I think we're stuck with it, to support those cases where the Number value is out of bounds :\

richardowright:
After some more thinking, I think it's pretty hopeless to go down this path and maintain backwards comparability. The above will break when doing any sort of comparison or looking up foreign keys which are now saved as string types.

My guess is a bit of documentation on the type handling would be best. That would at least give a better understanding of current behavior.

PR #144, opened 2018-07-26
chore: remove safe-buffer

JustinBeckwith:


ofrobots:
Shouldn't we be going the other way around? We need to have a safety net as it is very easy to use `Buffer` insecurely.

An alternative safety net would be to use the eslint rule [`no-buffer-constructor`](https://eslint.org/docs/rules/no-buffer-constructor) for JavaScript. I can't find a similar rule in tslint yet, but there might be a contrib rule [here](https://github.com/webschik/tslint-config-security) or this [feature request](https://github.com/Microsoft/tslint-microsoft-contrib/issues/249). In the meanwhile shouldn't we ban the use of `Buffer` directly?

JustinBeckwith:
@ofrobots we now have --throw-deprecation turned on for tests, and buffer constructors are deprecated in node 10.  Do you think that's enough?

ofrobots:
👍. That approach solves the Buffer constructor problem well.

Side note, generalizing: it seems like someone needs to go write a set of security oriented plugins for `tslint` as a counter part to [eslint-plugin-security](https://www.npmjs.com/package/eslint-plugin-security). Alternatively, perhaps we can run that plugin on the code generated by `tsc`?

JustinBeckwith:
Let me take that question one step further. Why does `tslint` need to exist? Wouldn't it make more sense to extend `eslint` with the ability to parse TypeScript, and apply a TypeScript plugin pack?

ofrobots:
Good question. `tslint` is more capable because it leans on the compiler for more accurate analysis. I would expect that eslint would 1) not want to depend on the compiler, 2) not want to reimplement typescript and 3) be generally less capable as a result.

ofrobots (approved):

PR #148, opened 2018-07-26
chore: drop dependency on common

JustinBeckwith:


src/index.js
```diff
@@ -364,10 +363,6 @@ function Datastore(options) {
     return new Datastore(options);
   }
 
-  options = common.util.normalizeArguments(this, options, {
```
ofrobots: Normalization of arguments is no longer necessary?

test/request.js
```diff
@@ -23,24 +23,31 @@ var is = require('is');
 var proxyquire = require('proxyquire');
 var sinon = require('sinon').sandbox.create();
 var through = require('through2');
-var util = require('@google-cloud/common').util;
+var pfy = require('@google-cloud/promisify');
```
ofrobots: style nit: cute but unnecessary abbreviation.

src/index.js
JustinBeckwith: I believe this was an artifact of the big `google-cloud` npm module.  @stephenplusplus to verify :)

stephenplusplus: Yep, that's right. All this did was extend the global configuration:

```js
var gcloud = require('gcloud')({ projectId: 'a' })

var datastore = gcloud.datastore()
// datastore.projectId = 'a'

var gcs = gcloud.storage({ projectId: 'b' })
// gcs.projectId = 'b'
```

PR #173, opened 2018-08-18
chore: start converting to es classes

JustinBeckwith:
@callmehiphop @stephenplusplus The system tests on master seem to fail for me (and this PR), but are passing in CI *scratches head*.  Can y'all try running the system tests against this branch?

There are actually very few changes in here, sorry the GH diff is so bad.


stephenplusplus:
Both master and this PR's system tests pass for me. Were you getting errors locally from this branch's system tests? What were they?

PR #176, opened 2018-08-21
chore: use arrow functions

JustinBeckwith:


gandazgul:
Can you use parenthesis always (around the params) to keep it consistent. There's parenthesis in some lines and not others.

PR #272, opened 2018-12-05
refactor: remove async module usage

vijay-qlogic:
Fixes #[2883](https://github.com/googleapis/google-cloud-node/issues/2883) partially for nodejs-datastore (it's a good idea to open an issue first for discussion)

- [ ] Tests and linter pass

### DIFF
--- a/system-test/datastore.ts
+++ b/system-test/datastore.ts
@@ -35,7 +35,6 @@ describe('Datastore', () => {
       const query = datastore.createQuery(kind).select('__key__');
 
       const [entities] = await datastore.runQuery(query);
-      // const entities = data[0];
       const keys = entities.map(entity => {
         return entity[datastore.KEY];
       });
--- a/system-test/datastore.ts
+++ b/system-test/datastore.ts
@@ -15,7 +15,6 @@
  */
 
 import * as assert from 'assert';
-import * as async from 'async';
 import {Datastore} from '../src';
 import {entity} from '../src/entity';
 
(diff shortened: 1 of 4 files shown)


system-test/datastore.ts
```diff
@@ -15,41 +15,39 @@
...
-          callback(err);
-          return;
-        }
-
-        const keys = entities.map(entity => {
-          return entity[datastore.KEY];
-        });
-
-        datastore.delete(keys, callback);
+      const data = await datastore.runQuery(query);
```
JustinBeckwith: collapse to `const [entities] = await datastore.runQuery(query);`

system-test/datastore.ts
```diff
@@ -15,41 +15,39 @@
...
     testKinds.push(keyObject.kind);
     return keyObject;
   };
 
-  after(done => {
-    function deleteEntities(kind, callback) {
-      const query = datastore.createQuery(kind).select('__key__');
+  after(async () => {
+    async function deleteEntities(kind) {
+      const query = await datastore.createQuery(kind).select('__key__');
```
JustinBeckwith: This doesn't need to be await'ed.  `createQuery` is a synchronous method.

system-test/datastore.ts
```diff
@@ -15,41 +15,39 @@
...
+      const entities = data[0];
+      const keys = entities.map(entity => {
+        return entity[datastore.KEY];
       });
+
+      await datastore.delete(keys);
     }
 
-    async.each(testKinds, deleteEntities, done);
+    let promises: Array<{}> = [];
```
JustinBeckwith: No need for the loop. 

`const promises = testKinds.map(kind => deleteEntities(kind))`

system-test/datastore.ts
```diff
@@ -861,34 +859,32 @@ describe('Datastore', () => {
               },
             ]);
 
-            transaction.commit(err => {
+            transaction.commit(async err => {
```
JustinBeckwith: Having an async callback method is super weird.  I believe `transaction.commit` is an async function.  Can you just `await` the result of that?

system-test/datastore.ts
```diff
@@ -861,34 +859,32 @@ describe('Datastore', () => {
...
-                  callback => {
-                    datastore.get(deleteKey, (err, entity) => {
-                      assert.ifError(err);
-                      assert.strictEqual(typeof entity, 'undefined');
-                      callback();
-                    });
-                  },
+              await Promise.all([
+                // Deletes the key that is in the deletion queue.
+                callback => {
```
JustinBeckwith: Mixing callback and promise based code is not good :)  These shouldn't be callbacks, and we should be using awaits throughout here.

system-test/datastore.ts
```diff
@@ -830,69 +824,51 @@ describe('Datastore', () => {
...
-                  callback => {
-                    datastore.get(deleteKey, (err, entity) => {
-                      assert.ifError(err);
-                      assert.strictEqual(typeof entity, 'undefined');
-                      callback();
-                    });
-                  },
+      await Promise.all([
+        // Deletes the key that is in the deletion queue.
+        datastore.get(deleteKey, (err, entity) => {
```
JustinBeckwith: Mixing callbacks and promises. Use a `.then` to do the assertion.

vijay-qlogic: Done! https://github.com/googleapis/nodejs-datastore/pull/272/commits/c120a83d74bdb5a4c96e361ab7cf40b4684839bc

JustinBeckwith:
👋 we could use a fixy on the merge conflicts :)

PR #305, opened 2019-01-19
refactor(types): enable noImplicitAny for transaction.ts & request.ts

### LINKED ISSUE #252: Enable `noImplicitAny` in the tsconfig
The goal here is to provide perfect TypeScript types for customers using the module.  Generally when doing this work I like to:

1. enable `noImplicitAny` in the tsconfig
2. Choose a single file to get passing, and fix all errors in that file (using VSCode here is really useful)
3. disable`noImplicitAny`
4. Get all the tests passing
5. Submit a PR for just the single file I updated
6. Repeat until finished

It's *a lot* easier if we do a PR per file on this one.

joecervino:
Fixes #252  (it's a good idea to open an issue first for discussion)

- [ ] Tests and linter pass


src/entity.ts
```diff
@@ -501,7 +501,7 @@ export namespace entity {
     };
 
     if (excludeFromIndexes && excludeFromIndexes.length > 0) {
-      excludeFromIndexes.forEach(excludePath => {
+      excludeFromIndexes.forEach((excludePath: string) => {
```
JustinBeckwith: Why do we need the type here?  What's the type of `excludeFromIndexes`?

src/request.ts
```diff
@@ -172,16 +171,13 @@ class DatastoreRequest {
...
    * });
    */
-  allocateIds(key, options, callback) {
-    if (entity.isKeyComplete(key)) {
-      throw new Error('An incomplete key should be provided.');
-    }
+  allocateIds(key: entity.Key, options: AllocateIdsOptions|number): Promise<google.datastore.v1.AllocateIdsResponse>;
+  allocateIds(key: entity.Key, options: AllocateIdsOptions|number, callback: AllocateIdsCallback): void;
+  allocateIds(key: entity.Key, options: AllocateIdsOptions|number, cb?: AllocateIdsCallback): void|Promise<google.datastore.v1.AllocateIdsResponse> {
+    if (entity.isKeyComplete(key)) throw new Error('An incomplete key should be provided.');
```
JustinBeckwith: I'm surprised the formatter allows this.  Can you add a newline and brackets to make this easier to read?

src/request.ts
```diff
@@ -172,16 +171,13 @@ class DatastoreRequest {
...
 
-    if (is.number(options)) {
-      options = {
-        allocations: options,
-      };
-    }
+    const callback = typeof cb === 'function' ? cb : cb!;
```
JustinBeckwith: This is confusing.  Isn't `cb` always a callback, unless its null/undefined? I think we could just not have this line of code.

src/request.ts
```diff
@@ -227,16 +223,14 @@ class DatastoreRequest {
    *     // All entities retrieved.
    *   });
    */
-  createReadStream(keys, options?) {
-    options = options || {};
+  createReadStream(key: Entities): Stream;
+  createReadStream(key: Entities, options?: CreateReadStreamOptions): Stream;
```
JustinBeckwith: You can safely delete this line.

src/request.ts
```diff
@@ -227,16 +223,14 @@ class DatastoreRequest {
...
+  createReadStream(keys: Entities, options: CreateReadStreamOptions = {}): Stream {
     keys = arrify(keys).map(entity.keyToKeyProto);
-    if (keys.length === 0) {
-      throw new Error('At least one Key object is required.');
-    }
+    if (keys.length === 0) throw new Error('At least one Key object is required.');
```
JustinBeckwith: same as above - we prefer the curly braces and new lines for readability pls.

src/request.ts
```diff
@@ -336,13 +330,12 @@ class DatastoreRequest {
...
    * });
    */
-  delete(keys, gaxOptions?, callback?) {
-    if (is.fn(gaxOptions)) {
-      callback = gaxOptions;
-      gaxOptions = {};
-    }
-
-    callback = callback || (() => {});
+  delete(keys:Entities): void;
```
JustinBeckwith: That doesn't seem right.  If no callback is provided, it should return a Promise<WhatEverType> here.

src/request.ts
```diff
@@ -336,13 +330,12 @@ class DatastoreRequest {
...
+  delete(keys: Entities, callback: google.datastore.v1.Datastore.CommitCallback): void;
+  delete(keys: Entities, gaxOptions: CallOptions, callback: google.datastore.v1.Datastore.CommitCallback): void;
+  delete(keys: Entities, gaxOptionsOrCallback?: CallOptions|google.datastore.v1.Datastore.CommitCallback, callback?: google.datastore.v1.Datastore.CommitCallback): void {
+    const gaxOptions = typeof gaxOptionsOrCallback === 'object' ? gaxOptionsOrCallback : {};
+    callback = typeof gaxOptionsOrCallback === 'function' ? gaxOptionsOrCallback : (() => {});
```
JustinBeckwith: Should be `: callback;`, not `(() => {})`

src/request.ts
```diff
@@ -452,19 +445,26 @@ class DatastoreRequest {
...
-      callback = options;
-      options = {};
-    }
-
-    options = options || {};
+  get(keys: Entities, options?: CreateReadStreamOptions): Promise<Entity|Stream>;
+  get(keys: Entities, callback: GetCallback): void;
+  get(keys: Entities, options: CreateReadStreamOptions, callback: GetCallback): void;
+  get(keys: Entities, optionsOrCallback?: CreateReadStreamOptions|GetCallback, cb?: GetCallback): void|Promise<Entity|Stream> {
+    if (typeof optionsOrCallback === 'function' && typeof optionsOrCallback !== 'object') {
```
JustinBeckwith: Well I mean both of these can't really be true? 
```js
const options = typeof optionsOrCallback === 'object' ? optionsOrCallback : {};
const callback = typeof optionsOrCallback === 'function' ? optionsOrCallback : cb;
```

src/request.ts
```diff
@@ -452,19 +445,26 @@ class DatastoreRequest {
...
+      cb = optionsOrCallback;
+      optionsOrCallback = {};
+    } else if (typeof optionsOrCallback === 'object' && optionsOrCallback !== null) {}
+    else optionsOrCallback = {};
+
+    let callback: GetCallback;
+    const options: CreateReadStreamOptions = optionsOrCallback;
+
+    if (typeof cb === 'function' && cb !== undefined) callback = cb;
```
JustinBeckwith: Always prefer more curly braces and new lines :)

src/request.ts
```diff
@@ -641,11 +643,11 @@ class DatastoreRequest {
    *     this.end();
    *   });
    */
-  runQueryStream(query: Query, options?) {
-    options = options || {};
+  runQueryStream(query: Query, options?: RunQueryStreamOptions): Stream;
```
JustinBeckwith: you can delete this overload

src/request.ts
```diff
@@ -938,17 +940,27 @@ class DatastoreRequest {
    *   const apiResponse = data[0];
    * });
    */
-  save(entities, gaxOptions?, callback?) {
+  save(entities: Entities, gaxOptions?: CallOptions): void|Promise<google.datastore.v1.ICommitResponse>;
```
JustinBeckwith: This should probably just be returning a promise, not void

src/request.ts
```diff
@@ -938,17 +940,27 @@ class DatastoreRequest {
...
+  save(entities: Entities, callback: SaveCallback): void;
+  save(entities: Entities, gaxOptionsOrCallback: CallOptions, callback?: SaveCallback): void;
+  save(entities: Entities, gaxOptionsOrCallback?: CallOptions|SaveCallback, cb?: SaveCallback): void|Promise<google.datastore.v1.ICommitResponse> {
     entities = arrify(entities);
 
-    if (is.fn(gaxOptions)) {
-      callback = gaxOptions;
-      gaxOptions = {};
-    }
+    if (typeof gaxOptionsOrCallback === 'function' && typeof gaxOptionsOrCallback !== 'object') {
```
JustinBeckwith: same feedback as above.  This logic is a little convoluted, and can be replaced with 2 lines

JCIV-SE:
@JustinBeckwith Thanks for taking the time out to look this over! I know it wasn't easy with so many fractured commits, I'm going to be making it a habit to keep it to one commit per issue. In general, a lot of the convoluted, unnecessary logic was created to pass the linter/formatter. I'll fix it asap

src/entity.ts
JCIV-SE: Array of strings on an Entity object which, at the moment, is set to type Entity. Interfaces in other parts of the codebase have `excludeFromIndexes` set to boolean. I'd like to make that more precise

src/request.ts
JCIV-SE: Done! Should I wrap things like `google.datastore.v1.AllocateIdsResponse` in extended interfaces to make them easier to read?

src/request.ts
JCIV-SE: Agreed! I added this line to get past the linter. It was throwing errors because we're trying to invoke a potentially undefined callback further down in the code

src/request.ts
JCIV-SE: Great!

src/request.ts
JCIV-SE: Agreed - another lint bypass for `[ts] Overload signature is not compatible with function implementation. [2394]` :) Changed!

src/request.ts
JCIV-SE: Another lint bypass for this error: `[ts]
Argument of type 'CommitCallback | undefined' is not assignable to parameter of type 'RequestCallback'.
  Type 'undefined' is not assignable to type 'RequestCallback'. [2345]` - Fixed!

src/request.ts
JCIV-SE: Nope! But the linter likes it 😆 Fixed!

src/request.ts
JCIV-SE: Roger that

src/request.ts
JCIV-SE: Great, was wondering about this one

src/request.ts
JCIV-SE: Okay!

src/request.ts
JCIV-SE: Done!

src/index.ts
```diff
@@ -34,6 +34,7 @@ import {entity} from './entity';
 import {Query} from './query';
 import {DatastoreRequest} from './request';
 import {Transaction} from './transaction';
+import {google} from '../proto/datastore';
```
JustinBeckwith: I'm a little surprised you need this, since it doesn't look like it's used anywhere.

src/request.ts
```diff
@@ -227,16 +228,15 @@ class DatastoreRequest {
    *     // All entities retrieved.
    *   });
    */
-  createReadStream(keys, options?) {
-    options = options || {};
+  createReadStream(key: Entities): Stream;
```
JustinBeckwith: You can just delete this line

src/request.ts
```diff
@@ -336,13 +336,17 @@ class DatastoreRequest {
...
-    callback = callback || (() => {});
+  delete(keys: Entities): Promise<google.datastore.v1.Datastore.CommitCallback>;
+  delete(keys: Entities, callback: google.datastore.v1.Datastore.CommitCallback): void;
+  delete(
+    keys: Entities, 
+    gaxOptions: CallOptions, 
+    callback: google.datastore.v1.Datastore.CommitCallback): void;
+  delete(keys: Entities, 
+    gaxOptionsOrCallback?: (CallOptions|google.datastore.v1.Datastore.CommitCallback), 
+    cb?: google.datastore.v1.Datastore.CommitCallback): void {
```
JustinBeckwith: The return type here should be `void|Promise<google.datastore.v1.Datastore.CommitCallback>`

src/request.ts
```diff
@@ -483,12 +490,14 @@ class DatastoreRequest {
    * @param {?error} callback.err An error returned while making this request
    * @param {object} callback.apiResponse The full API response.
    */
-  insert(entities, callback) {
+  insert(entities: Entities): Promise<google.datastore.v1.ICommitResponse>;
+  insert(entities: Entities, callback: CallOptions): void;
+  insert(entities: Entities, callback?: CallOptions): (void|Promise<google.datastore.v1.ICommitResponse>) {
```
JustinBeckwith: This is good, but you don't need the wrapping parens in `(void|Promise<google.datastore.v1.ICommitResponse>)`

src/request.ts
```diff
@@ -590,8 +599,8 @@ class DatastoreRequest {
       void;
   runQuery(query: Query, callback: RunQueryCallback): void;
   runQuery(
-      query: Query, optionsOrCallback?: RunQueryOptions|RunQueryCallback,
-      cb?: RunQueryCallback): void|Promise<RunQueryResponse> {
+      query: Query, optionsOrCallback?: (RunQueryOptions|RunQueryCallback),
```
JustinBeckwith: same feedback here and below, you don't need the wrappin parens

src/request.ts
```diff
@@ -938,17 +946,23 @@ class DatastoreRequest {
    *   const apiResponse = data[0];
    * });
    */
-  save(entities, gaxOptions?, callback?) {
+  save(entities: Entities, gaxOptions?: CallOptions): Promise<google.datastore.v1.ICommitResponse>;
+  save(entities: Entities, callback: SaveCallback): void;
+  save(
```
JustinBeckwith: I think you can delete this overload

src/request.ts
```diff
@@ -1111,12 +1128,12 @@ class DatastoreRequest {
    *
    * @private
    */
-  request_(config, callback) {
+  request_(config: RequestConfig): Promise<RequestResponse>;
+  request_(config: RequestConfig, callback: RequestCallback): void;
+  request_(config: RequestConfig, callback: RequestCallback = (() => {})): void|Promise<RequestResponse> {
```
JustinBeckwith: I don't think you want to default the callback to `(() => {}))`, and it should be marked as optional here.

JustinBeckwith: Oh wait!  This is the `request_` method.  I don't think we want the promise overloads here at all.  I *think* the promisification of this gets filtered out.

src/request.ts
```diff
@@ -1147,7 +1164,7 @@ class DatastoreRequest {
       };
     }
 
-    datastore.auth.getProjectId((err, projectId) => {
+    datastore.auth.getProjectId((err: Error|null|undefined, projectId: string|null|undefined) => {
```
JustinBeckwith: You really shouldn't need to explicitly define the types here.

src/transaction.ts
```diff
@@ -126,20 +128,29 @@ class Transaction extends DatastoreRequest {
...
    */
-  commit(gaxOptions?, callback?) {
-    if (is.fn(gaxOptions)) {
-      callback = gaxOptions;
-      gaxOptions = {};
+  commit(gaxOptions?: CallOptions): Promise<google.datastore.v1.CommitResponse>;
+  commit(callback: google.datastore.v1.Datastore.CommitCallback): void;
+  commit(gaxOptions: CallOptions, callback: google.datastore.v1.Datastore.CommitCallback): void;
+  commit(gaxOptionsOrCallback?: CallOptions|google.datastore.v1.Datastore.CommitCallback, cb?: google.datastore.v1.Datastore.CommitCallback): void|Promise<google.datastore.v1.CommitResponse> {
+    if (typeof gaxOptionsOrCallback === 'function' && typeof gaxOptionsOrCallback !== 'object') {
```
JustinBeckwith: this should use the simplified version of detection you use above

src/transaction.ts
```diff
@@ -360,19 +369,27 @@ class Transaction extends DatastoreRequest {
...
    * });
    */
-  rollback(gaxOptions, callback?) {
-    if (is.fn(gaxOptions)) {
-      callback = gaxOptions;
-      gaxOptions = {};
+  rollback(gaxOptions: CallOptions): Promise<google.datastore.v1.RollbackResponse>;
+  rollback(callback: google.datastore.v1.Datastore.RollbackCallback): void;
+  rollback(gaxOptionsOrCallback: CallOptions|google.datastore.v1.Datastore.RollbackCallback, cb?: google.datastore.v1.Datastore.RollbackCallback): void|Promise<google.datastore.v1.RollbackResponse> {
+    if (typeof gaxOptionsOrCallback === 'function' && typeof gaxOptionsOrCallback !== 'object') {
```
JustinBeckwith: use the simplified version

src/transaction.ts
```diff
@@ -430,14 +447,20 @@ class Transaction extends DatastoreRequest {
...
    */
-  run(options?, callback?) {
-    if (is.fn(options)) {
-      callback = options;
-      options = {};
+  run(options?: RunOptions): Promise<google.datastore.v1.BeginTransactionResponse>;
+  run(callback?: RunCallback): void;
+  run(options?: RunOptions, callback?: RunCallback): void;
+  run(optionsOrCallback?: RunOptions|RunCallback|Entity, cb?: RunCallback): void|Promise<google.datastore.v1.BeginTransactionResponse> {
+    if (typeof optionsOrCallback === 'function' && typeof optionsOrCallback !== 'object') {
```
JustinBeckwith: simplified version

system-test/datastore.ts
```diff
@@ -48,8 +48,10 @@ describe('Datastore', () => {
   it('should allocate IDs', done => {
     datastore.allocateIds(datastore.key('Kind'), 10, (err, keys) => {
       assert.ifError(err);
-      assert.strictEqual(keys.length, 10);
-      assert.strictEqual(entity.isKeyComplete(keys[0]), true);
+      if (keys !== null) {
```
JustinBeckwith: Wouldn't this be an error if `keys` was null?  Shouldn't this be `assert.ok(keys);`?

test/transaction.ts
```diff
@@ -18,6 +18,7 @@ import * as pfy from '@google-cloud/promisify';
 import * as arrify from 'arrify';
 import * as assert from 'assert';
 import * as proxyquire from 'proxyquire';
+import { configure } from 'protobufjs';
```
JustinBeckwith: this gets imported, but doesn't look like you use it

src/index.ts
JCIV-SE: Leftover from some testing, sorry about that!

src/request.ts
JCIV-SE: Sweet!

src/request.ts
JCIV-SE: Okay!

src/request.ts
JCIV-SE: Thanks, okay!

src/request.ts
JCIV-SE: Sounds good!

src/request.ts
JCIV-SE: Okay!

src/request.ts
JCIV-SE: Beautiful, I deleted the default callback and the promise overloads

src/request.ts
JCIV-SE: Agreed, it's just that I get an intense amount of errors if I don't. I'll change it and submit regardless

src/transaction.ts
JCIV-SE: 👍

src/transaction.ts
JCIV-SE: 👍

test/transaction.ts
JCIV-SE: Removed!

system-test/datastore.ts
JCIV-SE: Yup! That fixed it!

JustinBeckwith:
@JCIV-SE this looks like it needs a little love :) Is this going to get a look soon?

JCIV-SE:
@JustinBeckwith Yup! It's next on the list 👍

src/request.ts
```diff
@@ -336,13 +339,24 @@ class DatastoreRequest {
...
    */
-  delete(keys, gaxOptions?, callback?) {
-    if (is.fn(gaxOptions)) {
-      callback = gaxOptions;
-      gaxOptions = {};
-    }
-
-    callback = callback || (() => {});
+  delete(keys: Entities):
+      void|Promise<google.datastore.v1.Datastore.CommitCallback>;
```
JustinBeckwith: No need to explicitly return `void` here

src/request.ts
```diff
@@ -452,19 +466,24 @@ class DatastoreRequest {
...
    */
-  get(keys, options?): Promise<EntityDataObj[]>;
-  get(keys, options?, callback?): void|Promise<EntityDataObj[]> {
-    if (is.fn(options)) {
-      callback = options;
-      options = {};
-    }
-
-    options = options || {};
+  get(keys: Entities,
```
JustinBeckwith: Returning the Entity or a Stream doesn't feel quite right.  What am I missin here?

src/request.ts
```diff
@@ -452,19 +466,24 @@ class DatastoreRequest {
...
+      options?: CreateReadStreamOptions): Promise<Entity|Stream>;
+  get(keys: Entities, callback: GetCallback): void;
+  get(keys: Entities, options: CreateReadStreamOptions,
+      callback: GetCallback): void;
+  get(keys: Entities, optionsOrCallback?: CreateReadStreamOptions|GetCallback,
+      cb?: GetCallback): void|Promise<Entity|Stream> {
+    const options =
```
JustinBeckwith: We aren't doing this null check in other places, so I don't think it's entirely needed.

src/request.ts
```diff
@@ -483,7 +502,10 @@ class DatastoreRequest {
    * @param {?error} callback.err An error returned while making this request
    * @param {object} callback.apiResponse The full API response.
    */
-  insert(entities, callback) {
+  insert(entities: Entities): Promise<google.datastore.v1.ICommitResponse>;
+  insert(entities: Entities, callback: CallOptions): void;
```
JustinBeckwith: The type `CallOptions` for a callback doesn't seem right

src/request.ts
```diff
@@ -938,17 +959,21 @@ class DatastoreRequest {
    *   const apiResponse = data[0];
    * });
    */
-  save(entities, gaxOptions?, callback?) {
+  save(entities: Entities, gaxOptions?: CallOptions):
+      void|Promise<google.datastore.v1.ICommitResponse>;
```
JustinBeckwith: I don't think we need the void here

src/request.ts
```diff
@@ -1065,7 +1093,10 @@ class DatastoreRequest {
    * @param {?error} callback.err An error returned while making this request
    * @param {object} callback.apiResponse The full API response.
    */
-  update(entities, callback) {
+  update(entities: Entities): Promise<google.datastore.v1.ICommitResponse>;
+  update(entities: Entities, callback: CallOptions): void;
```
JustinBeckwith: `CallOptions` as types for callbacks is weird

test/transaction.ts
```diff
@@ -164,7 +164,7 @@ describe('Transaction', () => {
       const gaxOptions = {};
 
       transaction.request_ = config => {
-        assert.strictEqual(config.gaxOpts, gaxOptions);
```
JustinBeckwith: Please use `assert.deepStrictEqual(config.gaxOpts, gaxOptions)`

src/request.ts
joecervino: Agreed, but I added the void to solve this error:
```
Property 'delete' in type 'Transaction' is not assignable to the same property in base type 'DatastoreRequest'.
  Type '(entities: any) => void' is not assignable to type '{ (keys: any): Promise<CommitCallback>; (keys: any, callback: CommitCallback): void; (keys: any, gaxOptions: CallOptions, callback: CommitCallback): void; }'.
    Type 'void' is not assignable to type 'Promise<CommitCallback>'.
```
delete in transaction.ts has a void return annotation, and when changed to the same Promise<CommitCallback> there's an error asking explicitly for an any or void return annotation. I kept the promise to follow the "no callback = return promise" rule.

Should I make them all void and remove the promise return annotation from the method definition? (if all method overloads have the same return annotation, do we really need the overloads?)

JustinBeckwith: Well hold on nah - what does `CommitCallback` have in it's type?  Is it a callback function type?  The answer here may be to return `Promise<void>`.

test/transaction.ts
joecervino: Great! Okay!

src/request.ts
joecervino: Getting the same error/problem as `delete` - I'll figure something out

src/request.ts
joecervino: Yeah I agree, I'd prefer the consistency - the problem is that `typeof null` evalutes to `'object'` in JS (thank you C) and a test uses null to trigger options be set to an object literal:
```
it('should default options to an object', done => {
        request.get(keys, null, err => {  // 2nd arg 'null' is what's being used to trigger the options object correction logic
          assert.ifError(err);

          const spy = request.createReadStream.getCall(0);
          assert.deepStrictEqual(spy.args[1], {});
          done();
        });
      });
```
Was thinking we should actually add the null check in other places because this doesn't tend to be accounted for

src/request.ts
joecervino: Yeah, `CommitCallback` is a callback function type. I also tried `Promise<void>` but I'm getting the same errors. I'll ask around to see if we can come up with something using feedback from an IDE

joecervino: The error I'm solving for:
```
Type 'void' must have a '[Symbol.iterator]()' method that returnsan iterator.

826       const [[deletedEntity], [fetchedEntity]] = await Promise.all([
```

src/request.ts
joecervino: I admit the reasoning isn't 100% here. `get` relies on `createReadStream` which returns a stream, but `get` is expecting something iterator-related:
```
Type 'Stream' must have a '[Symbol.iterator]()' method that returns an iterator.

826       const [[deletedEntity], [fetchedEntity]] = await Promise.all([
```
Changing it to `Promise<Entity>` works well - lmk your take on it :)

PR #310, opened 2019-01-23
fix(types): Make gaxOptions optional in Transaction.rollback()

### LINKED ISSUE #309: gaxOptions parameter is not optional in rollback()
By analogy with `Transaction.commit()`
https://github.com/googleapis/nodejs-datastore/blob/4bbe64ff09e28ffac063714def174860388f136c/src/transaction.ts#L129
the first parameter of `rollback()`, `gaxOptions`, should be optional:
https://github.com/googleapis/nodejs-datastore/blob/4bbe64ff09e28ffac063714def174860388f136c/src/transaction.ts#L363

evgenius:
Fixes #309.

- [X] Tests and linter pass
- [X] Code coverage does not decrease (if any source code was changed)

### DIFF
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -360,7 +360,7 @@ class Transaction extends DatastoreRequest {
    *   const apiResponse = data[0];
    * });
    */
-  rollback(gaxOptions, callback?) {
+  rollback(gaxOptions?, callback?) {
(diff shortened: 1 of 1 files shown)


evgenius:
I signed it!

PR #344, opened 2019-03-01
refactor(ts): enable noImplicitAny on test/transaction.ts

### LINKED ISSUE #326: Enable noImplicitAny in the tsconfig
We're at a point where the types are pretty good in the source files, it's mostly the tests that are still messed up.  Lets close this TypeScript thing out :)

joecervino:
Fixes #326 

- [ ] Tests and linter pass


src/transaction.ts
```diff
@@ -641,14 +630,14 @@ class Transaction extends DatastoreRequest {
   }
 }
 
+//tslint:disable-next-line no-any
+type Any = any;
```
JustinBeckwith: Where is this used!?

test/transaction.ts
```diff
@@ -18,12 +18,21 @@ import * as pfy from '@google-cloud/promisify';
 import * as arrify from 'arrify';
 import * as assert from 'assert';
 import * as proxyquire from 'proxyquire';
+import { google } from '../proto/datastore';
+import { RequestConfig } from '../src/request';
+import { Entity } from '../src/entity';
+import { Transaction } from '../src';
+
+//tslint:disable-next-line no-any
+type Any = any;
```
JustinBeckwith: Not a big fan of this.  It's basically just disabling `noImplicitAny`.

test/transaction.ts
```diff
@@ -18,12 +18,21 @@ import * as pfy from '@google-cloud/promisify';
...
+type Fn = (...args: Any) => void;
```
JustinBeckwith: This exists, it's just called `Function` 😛

test/transaction.ts
```diff
@@ -32,8 +41,8 @@ const fakePfy = Object.assign({}, pfy, {
   },
 });
 
-// tslint:disable-next-line no-any variable-name
-const DatastoreRequestOverride: any = {
+// tslint:disable-next-line variable-name
+const DatastoreRequestOverride: Any = {
```
JustinBeckwith: Instead of `any` here, can we do ` as {} as DatastoreRequest` or something to that effect?

test/transaction.ts
```diff
@@ -56,8 +65,8 @@ class FakeDatastoreRequest {
 
 describe('Transaction', () => {
   // tslint:disable-next-line variable-name
-  let Transaction;
-  let transaction;
+  let Transaction: Any;
```
JustinBeckwith: Should be something like `: typeof tsTypes.Transaction`

test/transaction.ts
```diff
@@ -56,8 +65,8 @@ class FakeDatastoreRequest {
...
+  let transaction: Any;
```
JustinBeckwith: Should be `Transaction`

test/transaction.ts
```diff
@@ -151,7 +159,7 @@ describe('Transaction', () => {
     });
 
     it('should commit', done => {
-      transaction.request_ = config => {
+      transaction.request_ = (config: RequestConfig & {gaxOptions: unknown}) => {
```
JustinBeckwith: while we're in here, maybe use `sandbox.stub` for this so we don't need to restate the types?

test/transaction.ts
```diff
@@ -163,7 +171,7 @@ describe('Transaction', () => {
     it('should accept gaxOptions', done => {
       const gaxOptions = {};
 
-      transaction.request_ = config => {
+      transaction.request_ = (config: RequestConfig) => {
```
JustinBeckwith: Here and below - we want `transaction` to be properly typed, as called out above.  That way, you can change this to `sandbox.stub` and have it retain typing for you, so we don't need to say the type twice.

test/transaction.ts
```diff
@@ -188,17 +196,17 @@ describe('Transaction', () => {
       const rollbackApiResponse = {};
 
       beforeEach(() => {
-        transaction.rollback = (callback) => {
+        transaction.rollback = (callback: Fn) => {
```
JustinBeckwith: Just use `Function` :)

test/transaction.ts
joecervino: I was originally using `Function`, but I was like "Maybe someone will want to see what this means" 🤦‍♂️

Fixed!

test/transaction.ts
joecervino: This one was getting me a bit - in case I'm not understanding you correctly, I wrote:
```
import * as tsTypes from '../src/transaction'
...
let Transaction: typeof tsTypes.Transaction;
let transaction: tsTypes.Transaction
```

JustinBeckwith: PERFECT.

test/transaction.ts
joecervino: Got it 👍

test/transaction.ts
```diff
@@ -56,8 +65,8 @@ class FakeDatastoreRequest {
```
joecervino: @JustinBeckwith running into a slight bug that could use your quick input - on above lines [54] and [61] and below on line [191], we reassign the `save`, `delete`, and `rollback` request.ts methods to signatures that are being rejected for incompatibility

Basically, `delete` and `save` can't be reassigned to a signature that returns `void` because both are supposed to return promises when no callback is passed 
(Save: https://github.com/googleapis/nodejs-datastore/blob/master/src/request.ts#L962  
Delete: https://github.com/googleapis/nodejs-datastore/blob/master/src/request.ts#L342)

I can hack my way through and make the overloads return `void` instead of a promise, but that wouldn't be correct - was wondering what you'd recommend?

JustinBeckwith: Gah - yeah that can suck with the signature bits. I am ok with whatever casting you need in that situation.  I *do* like using sinon for the stubbing though, because I like having the option to check the `stub.calledOnce` stuff if I need to, and the security of knowing `sandbox.restore()` will undo all my shenanigans.

joecervino: I managed a casting approach, but I'm trying to make use of sinon for you (which means I have to learn Sinon again) - wanted to explain the delay

JustinBeckwith:
@joecervino looking better!  Can you `npm run fix` to get CI happy?  Then I'll take a look.

src/request.ts
```diff
@@ -258,7 +258,7 @@ class DatastoreRequest {
             reqOpts,
             gaxOpts: options.gaxOptions,
           },
-          (err: Error, resp: Entity) => {
```
JustinBeckwith: Just drop the types here entirely, as they can be inferred.

src/request.ts
```diff
@@ -339,20 +339,14 @@ class DatastoreRequest {
...
    */
-  delete(keys: Entities):
-      void|Promise<google.datastore.v1.Datastore.CommitCallback>;
-  delete(
-      keys: Entities,
-      callback: google.datastore.v1.Datastore.CommitCallback): void;
-  delete(
-      keys: Entities, gaxOptions: CallOptions,
-      callback: google.datastore.v1.Datastore.CommitCallback): void;
+  delete(): Promise<google.datastore.v1.CommitResponse>;
```
JustinBeckwith: That isn't the right type.  All of the promises for our APIs resolve with `Promise<[SomeType, OriginalProtoBuf]>` or something like that.  This should probably be `Promise<[google.datastore.v1.CommitResponse]>`.  That thing is long enough that I would create:
```js
export type CommitResponse = [google.datastore.v1.Datastore.CommitCallback];
```

And then just use that in the other signatures.

src/request.ts
```diff
@@ -959,6 +953,7 @@ class DatastoreRequest {
    *   const apiResponse = data[0];
    * });
    */
+  save(entities: Entities): void;
```
JustinBeckwith: I don't think this is right.  We want `save` to return a `Promise<CommitResponse>` here.  First issue is the `void`, second is the array thing pointed out above.

src/request.ts
joecervino: Yikes, yeah I kind of knew this wouldn't fly, was trying to be space-efficient. Fixed  👍

src/request.ts
joecervino: 🤦‍♂️ gotcha, sorry about that

joecervino:
@JustinBeckwith just a friendly ping, changes made!

src/request.ts
JustinBeckwith: Still don't think this type is right.  Probably should be `Promise<CommitResponse>`

src/request.ts
```diff
@@ -339,20 +339,14 @@ class DatastoreRequest {
...
+  delete(keys: Entities, gaxOptions: CallOptions, callback: CommitCallback):
+      void;
   delete(
-      keys: Entities,
-      gaxOptionsOrCallback?: CallOptions|
-      google.datastore.v1.Datastore.CommitCallback,
-      cb?: google.datastore.v1.Datastore.CommitCallback):
-      void|Promise<google.datastore.v1.Datastore.CommitCallback> {
+      keys?: Entities, gaxOptionsOrCallback?: CallOptions|CommitCallback,
+      cb?: CommitCallback): void|Promise<google.datastore.v1.CommitResponse> {
```
JustinBeckwith: Same as above, promise responses come in an array

src/request.ts
```diff
@@ -1044,23 +1038,22 @@ class DatastoreRequest {
...
         callback(err, resp);
         return;
       }
 
-      arrify(resp.mutationResults)
-          .forEach((result: Entity, index: number) => {  //! Entity malo
-            if (!result.key) {
-              return;
-            }
+      arrify(resp.mutationResults).forEach((result: Entity, index: number) => {
```
JustinBeckwith: why do we need explicit types here?  Can we dump `: Entity` and `:number`

src/request.ts
```diff
@@ -959,14 +953,14 @@ class DatastoreRequest {
...
-      Promise<google.datastore.v1.ICommitResponse>;
+  save(entities: Entities): Promise<CommitResponse>;
+  save(entities: Entities, gaxOptions?: CallOptions): Promise<CommitResponse>;
   save(entities: Entities, gaxOptions: CallOptions, callback: SaveCallback):
       void;
   save(entities: Entities, callback: SaveCallback): void;
   save(
       entities: Entities, gaxOptionsOrCallback?: CallOptions|SaveCallback,
-      cb?: SaveCallback): void|Promise<google.datastore.v1.ICommitResponse> {
+      cb?: SaveCallback): void|Promise<CommitResponse> {
```
JustinBeckwith: Here and all instances below

src/transaction.ts
```diff
@@ -332,7 +329,10 @@ class Transaction extends DatastoreRequest {
    *   });
    * });
    */
-  delete(entities: Entity): void {
+  delete(): Promise<google.datastore.v1.CommitResponse>;
+  delete(entities: Entities): void;
+  delete(entities?: Entities):
+      void|Promise<google.datastore.v1.CommitResponse> {
```
JustinBeckwith: should be `CommitResponse`

src/transaction.ts
```diff
@@ -378,12 +378,17 @@ class Transaction extends DatastoreRequest {
...
-  rollback(gaxOptions?: CallOptions): Promise<RollbackResponse>;
-  rollback(callback: RollbackCallback): void;
-  rollback(gaxOptions: CallOptions, callback: RollbackCallback): void;
+  rollback(): void;
+  rollback(callback: google.datastore.v1.Datastore.RollbackCallback): void;
+  rollback(gaxOptions: CallOptions):
+      Promise<google.datastore.v1.RollbackResponse>;
+  rollback(
+      gaxOptions: CallOptions,
+      callback: google.datastore.v1.Datastore.RollbackCallback): void;
```
JustinBeckwith: Just `RollbackCallback`?

test/transaction.ts
```diff
@@ -188,21 +198,22 @@ describe('Transaction', () => {
...
       });
 
       it('should pass the commit error to the callback', done => {
-        transaction.commit((err, resp) => {
-          assert.strictEqual(err, error);
-          assert.strictEqual(resp, apiResponse);
-          done();
-        });
+        transaction.commit(
+            (err: Error|null, resp?: google.datastore.v1.CommitResponse) => {
```
JustinBeckwith: Why the types?

test/transaction.ts
```diff
@@ -211,11 +222,13 @@ describe('Transaction', () => {
       transaction.request_ = (config, callback) => {
         callback(null, resp);
       };
-      transaction.commit((err, apiResponse) => {
-        assert.ifError(err);
-        assert.deepStrictEqual(resp, apiResponse);
-        done();
-      });
+      transaction.commit(
+          (err: Error|null,
```
JustinBeckwith: why the types?

src/request.ts
joecervino: Weird, I thought I had changed all of these over. Sorry about that Justin :l

test/transaction.ts
joecervino: Should we keep adding `tslint:disable-next-line no-any` whenever we use `any` then? Just thought it'd clean things up a bit 🤷‍♂️

src/request.ts
joecervino: `number` yes, `result` is implicitly being assigned `any` 😅

JustinBeckwith:
Please run `npm run fix`

joecervino:
@JustinBeckwith done :)

Thanks btw, I appreciate you taking the time to get through this

PR #346, opened 2019-03-04
fix: add missing package, and add install test

JustinBeckwith:
Ok, so good news / bad news on this one.  I added an install test to exercise the TypeScript package, and ensure baseline compilation.  That uncovered a bug with `duplexify` types, so I just added that package for now.  It did *not* however uncover any issues with a missing long package.


package.json
```diff
@@ -47,6 +47,7 @@
   "dependencies": {
     "@google-cloud/projectify": "^0.3.2",
     "@google-cloud/promisify": "^0.4.0",
+    "@types/duplexify": "^3.6.0",
```
callmehiphop: should this be a dev dependency?

JustinBeckwith: *sighs* no.  That's the error this uncovered.  `gax` is returning this somewhere in a class that is being *extended* here.

callmehiphop: That sounds .. weird? Sorry if I'm misunderstanding, buuut would it better if we just made a change that casts said stream as a plain old Duplex?

JustinBeckwith: It's possible that's the right change *upstream* in `gax`.  The real problem here is that gax leaks types from other packages.  The "correct" fix is probably to make this change in *gax* instead of here. 

*scratches head*

Thoughts?

callmehiphop: SGTM!

PR #351, opened 2019-03-06
fix: include types in package

### LINKED ISSUE #350: package.json "types: src/build/index.d.ts" are missing
I want to use the library in Typescript and consume its types. But `package.json` currently lacks `types: "src/build/index.d.ts` field that will instruct Typescript to use these types. Without it it doesn't matter that you ship `*.d.ts` files - they are not being used in typescript.

Fix should be fairly simple - just add this missing field to `package.json`.

I know it can default to "index.d.ts", but only if it's in root directory. In your case - it's under `build/src/`, that's why it doesn't work by default.

So far was relying on unstable third-party types from `@types/google-cloud__datastore`, cannot wait to start using official types:)

JustinBeckwith:
This leaves me with so many questions, but here I am.  Fixes #350

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "build/protos",
     "build/src"
   ],
+  "types": "build/src/index.d.ts",
   "keywords": [
     "google apis client",
     "google api client",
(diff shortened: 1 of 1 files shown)


JustinBeckwith:
TBR = @stephenplusplus

PR #352, opened 2019-03-06
fix: make an explicit dependency on long

### LINKED ISSUE #345: Missing explicit dependency on Long.js
The `datastore.d.ts` and `entity.d.ts` files that are included in the build uses Long.js. There are several type definitions there that use the `Long` type.

Long.js is not declared in package.json as a direct dependency of  `nodejs-datastore`, which causes compilation to fail in some monorepo situations where dependencies may get hoisted. In our case, using yarn.

It works as expected without monorepos, because long.js is a transitive dependency from protobuf.js

  - `@google-cloud/datastore` version: 3.1.1

#### Steps to reproduce

Sorry, wasn't able to reproduce outside of our monorepo, which I'm not able to share.

Cheers

JustinBeckwith:
I learned something today!  The `-i` flag on `pbts` will let you specific which imports to explicitly bring in.  Not sure why I need to do this, but 🤷‍♂️   Fixes #345.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "prepare": "npm run compile",
     "pretest": "npm run compile",
     "proto": "npm run proto:datastore",
-    "proto:datastore": "mkdir -p proto && pbjs -t static-module -w commonjs -p node_modules/google-proto-files google/datastore/v1/datastore.proto | pbts -o proto/datastore.d.ts -",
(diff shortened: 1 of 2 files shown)


callmehiphop:
Oh man, I've failed you @JustinBeckwith looks like this import `long` and the global reference is `Long`. I just noticed when making the same change in pubsub 😭

JustinBeckwith:
This was like... not an obvious fix.

JustinBeckwith:
Does this mean I didn't fix it?

callmehiphop:
Yeah, I think so, lol.

SimenB:
I opened up #358. I _think_ that's it 🙂

PR #356, opened 2019-03-07
fix(docs): move ts overloads above doc string

### LINKED ISSUE #348: Docs for runQuery and getin  v3.0
#### Environment details
  - `@google-cloud/datastore` version: 3.0

#### Steps to reproduce
The new docs about datastore v3.0 are missing a few things:

- runQuery (in 2.0.x docs we had them available [here](https://cloud.google.com/nodejs/docs/reference/datastore/2.0.x/DatastoreRequest#runQuery). In the 3.0 there is only [runQueryStream](https://cloud.google.com/nodejs/docs/reference/datastore/3.0.x/DatastoreRequest#runQueryStream). This functions was not removed from v2 to v3, cause i checked in the code and its still there.
- get (in the 2.0 docs was [here](https://cloud.google.com/nodejs/docs/reference/datastore/2.0.x/DatastoreRequest#get). In the 3.0 docs is missing
- Same problem with get and runQuery in the Datastore class too. In the 3.0 docs they are missing. [Example](https://cloud.google.com/nodejs/docs/reference/datastore/2.0.x/Datastore#get)

callmehiphop:
Fixes #348

- [x] Tests and linter pass

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -615,6 +615,10 @@ class Datastore extends DatastoreRequest {
   static NO_MORE_RESULTS = 'NO_MORE_RESULTS';
   NO_MORE_RESULTS = Datastore.NO_MORE_RESULTS;
 
+  createQuery(kind?: string): Query;
+  createQuery(kind?: string[]): Query;
+  createQuery(namespace: string, kind: string): Query;
+  createQuery(namespace: string, kind: string[]): Query;
   /**
    * Create a query for the specified kind. See {@link Query} for all
    * of the available methods.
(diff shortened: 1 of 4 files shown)


stephenplusplus:
Maybe our docs test command should throw if a method doesn't have any docs. Would that have caught this?

callmehiphop:
@stephenplusplus well our docs test only checks for broken links AFAIK. Although I think it would be awesome if we expanded our doc test task to do that and maybe lint with something like [write-good](https://www.npmjs.com/package/write-good).

PR #358, opened 2019-03-08
fix: correct Long types import

### LINKED ISSUE #345: Missing explicit dependency on Long.js

SimenB:
Fixes #345

Ref #352


SimenB:
ergh no, the generator emoves the manual edit I made 😅 Does need the dependency in package.json, though

package.json
```diff
@@ -49,6 +49,7 @@
     "@google-cloud/projectify": "^0.3.2",
     "@google-cloud/promisify": "^0.4.0",
     "@types/duplexify": "^3.6.0",
+    "@types/long": "^4.0.0",
```
JustinBeckwith: you don't need this because it's shipped with `protobufjs` :)

proto/datastore.d.ts
```diff
@@ -1,5 +1,5 @@
 import * as $protobuf from "protobufjs";
-import * as long from "long";
+import * as Long from "long";
```
JustinBeckwith: Yarrr, nice find and sorry for the mixup.  What you *really* wanna do here is update this line in package.json:

```
"proto:datastore": "mkdir -p proto && pbjs -t static-module -w commonjs -p node_modules/google-proto-files google/datastore/v1/datastore.proto | pbts -i long -o proto/datastore.d.ts -",
```

Make that:
```
"proto:datastore": "mkdir -p proto && pbjs -t static-module -w commonjs -p node_modules/google-proto-files google/datastore/v1/datastore.proto | pbts -i Long -o proto/datastore.d.ts -",
```

That way the fix doesn't get roughed up by the generator.

SimenB: Then it's `import Long from 'Long'` which I guess will go boom on linux?

package.json
SimenB: Yeah, but you shouldn't depend on it existing transitively. It won't work with `yarn pnp` for instance

proto/datastore.d.ts
JustinBeckwith: Nah, it'll probably be fine

package.json
JustinBeckwith: ah, cool - did not know that!

proto/datastore.d.ts
mgabeler-lee-6rs: This very much will blow up on Linux -- import filenames are case sensitive there.  Applying the suggestion results in:
```
> @google-cloud/datastore@3.1.2 compile /home/mgl/3psrc/nodejs-datastore
> tsc -p . && cp -r src/v1 build/src && cp -r proto* build && cp test/*.js build/test

proto/datastore.d.ts:2:23 - error TS2307: Cannot find module 'Long'.

2 import * as Long from "Long";
                        ~~~~~~


Found 2 errors.
```

stephenplusplus: Is this what we want? https://github.com/dcodeIO/protobuf.js/pull/1166

JustinBeckwith: Yes lol.  That would fix all the things.

callmehiphop:
@JustinBeckwith in some of the other clients we are just editing the generated file until a better solution presents itself. Assuming we're cool doing that here as well, I think this PR is good to go unless you have any objections

JustinBeckwith:
SGTM.  Let's just be careful about not blasting these away :)

callmehiphop:
@JustinBeckwith I see you became a contributor to Protobuf.js 😮 any chance we'll see a release soon with the Long fix?

PR #435, opened 2019-06-12
fix: assign more restrictive type to PathType (#433)

### LINKED ISSUE #433: datastore.key(['City', undefined]) should be a TypeScript error
#### Environment details

  - OS: Darwin (macOS 10.14.5)
  - Node.js version: v10.15.1
  - npm version: 6.9.0
  - `@google-cloud/datastore` version: 4.0.0

#### Steps to reproduce

  1. Create a key with `undefined` in its path, `datastore.key(['City', undefined])`
  2. Compile and run the code.

Actual result: runtime error:
```
Error: Key path element must not be incomplete: [City: ]
```

Expected result: compile-time error.

growtofill:
This prevents `Key path element must not be incomplete` runtime error, and also allows using big int
(datastore.Int) when creating a key using KeyOptions object as an argument.

Fixes #433 (it's a good idea to open an issue first for discussion)

- [x] Tests and linter pass

### DIFF
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -18,6 +18,7 @@ import arrify = require('arrify');
 import * as extend from 'extend';
 import * as is from 'is';
 import {Query, QueryProto} from './query';
+import {PathType} from '.';
 
 // tslint:disable-next-line no-namespace
 export namespace entity {
@@ -162,7 +163,7 @@ export namespace entity {
 
   export interface KeyOptions {
     namespace?: string;
-    path: Array<string | number>;
+    path: PathType[];
   }
 
   /**
--- a/src/index.ts
+++ b/src/index.ts
@@ -37,8 +37,7 @@ import {Transaction} from './transaction';
 
 const {grpc} = new GrpcClient();
 
-// tslint:disable-next-line: no-any
-export type PathType = any;
+export type PathType = string | number | entity.Int;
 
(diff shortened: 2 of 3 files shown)


growtofill:
As a Google TVC, I shouldn't need to sign the CLA explicitly, let me check with my manager.

callmehiphop:
Is this a breaking change for TypeScript users? If so we should update the commit message to indicate that

growtofill:
@callmehiphop yes, in a sense that the code that use to compile may not compile after this update. Is there an example of such message I can use for a reference?

JustinBeckwith:
I think it's fine to call that out in the notes - but I don't think this should be a semver major bump.  This is a "fix" in that it's actually enforcing the types are adequately limited.  Any concerns with including this in a patch release?

callmehiphop:
Applying semver to fix types has always been pretty fuzzy for me, I think in the case of transitioning away from `any` I get a little nervous since it's much more drastic of a change than fixing an incorrect type or adding a missing property. Buuut, if you guys think it's the right way to go, then I say let's go for it! 😎

JustinBeckwith:
@growtofill have you been able to work out the CLA issues?  Would love to merge this!

growtofill:
@JustinBeckwith updated to use private email, sorry for the delay.

PR #436, opened 2019-06-12
fix: add `parseInt` to constructor of task keys

BenWhitehead:
The `id` property of `Key` is a string, update the use site of the key
to parse it as an int when being used to create a key.


BenWhitehead:
> I think we want to use `datastore.int(taskId)` instead of `parseInt`. Technically both will work, but it would be great to include some of the built in helpers methods into a sample (they sometimes get missed)

Thanks, I didn't know that function existed. Pushed a new commit using `datastore.int`.

PR #439, opened 2019-06-13
fix(docs): fix <caption> tag rendering in Key doc

jkwlui:
`<caption>` tag is being rendered out inside the code block. 

[link](https://googleapis.dev/nodejs/datastore/latest/Key.html)

### DIFF
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -174,21 +174,28 @@ export namespace entity {
    * @param {string} [options.namespace] Optional namespace.
    *
    * @example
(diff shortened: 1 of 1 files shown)


callmehiphop:
@jkwlui I think the reason is that `<caption>` tags have to be on the same line as the `@example` tag. Anything that comes in after on a newline is considered the example

JustinBeckwith (approved):

bcoe (approved):
a little late to the party :stuck_out_tongue:

PR #450, opened 2019-06-27
build: use config file for linkinator

JustinBeckwith:


linkinator.config.json
```diff
@@ -0,0 +1,7 @@
+{
+  "recurse": true,
+  "skip": [
+    "https://codecov.io/gh/googleapis/",
```
bcoe: this is great.

PR #451, opened 2019-06-28
feat: excludeFromIndex allow for '*' wildcard to catch all properties on object

praveenqlogic01:
towards #12 

- [ ] Tests and linter pass


src/request.ts
```diff
@@ -1040,6 +1040,13 @@ class DatastoreRequest {
           }
         }
 
+        if (entityObject.excludeFromIndexes) {
```
stephenplusplus: We have the 'excludePathFromEntity' function in entity.ts. Could this new behavior be put into the same function? This is probably the most complex bit of code in our library, so anything we can do to make it more approachable is great.

praveenqlogic01: @stephenplusplus Hi, as suggested i implemented the new behavior in  function  excludePathFromEntity!

jacekkopecky:
Thank you, this will be oh so useful!

Is there documentation anywhere?

It seems that `prop.*` excludes everything under the property `prop`; and there's a hint that a simple `.*` excludes everything. Is that correct?

If so, shouldn't that be just `*` instead? Why the initial dot?

system-test/datastore.ts
```diff
@@ -129,6 +129,69 @@ describe('Datastore', () => {
       await datastore.delete(postKey);
     });
 
+    it('should remove index with using wildcrad in excludeFromIndexes', async () => {
```
jacekkopecky: typo `wildcrat` -> `wildcard`

praveenqlogic01: done!

praveenqlogic01:
> Thank you, this will be oh so useful!
> 
> Is there documentation anywhere?
> 
> It seems that `prop.*` excludes everything under the property `prop`; and there's a hint that a simple `.*` excludes everything. Is that correct?
> 
> If so, shouldn't that be just `*` instead? Why the initial dot?

@jacekkopecky Hi, yes that's right, "." is representing root path here and * representing everything from there!

praveenqlogic01:
verifying cla done!

uschtwill:
This isn't documented at all, is it?

test/request.ts
```diff
@@ -1392,6 +1392,158 @@ describe('Request', () => {
...
+          excludeFromIndexes
+        );
+        done();
+      };
+
+      request.save(
+        {
+          key,
+          data,
+          excludeFromIndexes: [
```
siddjain, line 1535: how can i exclude all properties from being indexed? will ['*'] do it?

kirillgroshkov:
Is there documentation for this except source code of this pull request?

PR #452, opened 2019-07-03
feat: add merge method for merging an object into an existing entity

### LINKED ISSUE #10: datastore.save() with upsert should allow get-or-create
###### Copied from original issue: https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2419

<a href="/jgeewax"></a><a href="/jgeewax">@&shy;jgeewax</a><br>June 29, 2017 10:55 PM

Ideally, datastore's upsert would be one of those "get or create" type methods, but I can't figure out how to make it act like get or create...

That is, what I want to do is:

```js
function getOrCreate(id, name) {
  var entity = {
    key: datastore.key(['User', id]),
    method: 'upsert',
    data: {
      name: name
    }
  };
  return datastore.save(entity).then( function (data) {
    return data[0][0]; // Or whatever to return the entity as it actually exists in datastore
  });
```

Turns out that the data coming back in the promise is ... some mutation results?

```js
datastore.save(entity).then( function (data) {
  console.log(data[0].mutationResults[0]);
});
```

spits out:

```
{ key: null,
  version: '1498776439005000',
  conflictDetected: false }
```

Which is lovely to know, but doesn't really help me at all.

Can we please make this method actually useful to people who want to use get-or-create? (For comparison, Datastore from App Engine can do this in Python...) Also, let me know if I'm crazy and just using this wrong. The docs aren't too descriptive about the output from the promises when you specify the method to use in the save request.

/cc @pcostell @stephenplusplus

praveenqlogic01:
Fixes #10 

- [ ] Tests and linter pass


src/request.ts
```diff
@@ -1186,6 +1187,48 @@ class DatastoreRequest {
...
+   * @param {object} entities.data Data to merge to the same for provided key.
+   * @param {function} callback The callback function.
+   * @param {?error} callback.err An error returned while making this request
+   * @param {object} callback.apiResponse The full API response.
+   */
+  merge(
+    entities: Entities,
+    callback?: CallOptions
+  ): void | Promise<CommitResponse> {
+    arrify(entities)
```
stephenplusplus: I believe we should have this in a Transaction, so that the entity is locked while we are getting/updating.

stephenplusplus: Here's an example: https://github.com/googleapis/nodejs-datastore/blob/6b9ce7712af06f4a5e9c2bc2cc9f8c3ef31f2aa5/samples/concepts.js#L1080-L1100

src/request.ts
```diff
@@ -1186,6 +1187,48 @@ class DatastoreRequest {
     this.save(entities, callback);
   }
 
+  merge(entities: Entities): Promise<CommitResponse>;
+  merge(entities: Entities, callback: CallOptions): void;
+  /**
+   * Maps to {@link Datastore#save}, forcing the method to be `merge`.
```
stephenplusplus: Can you beef up the docs for this method? It should be mentioned how the process works and how it is accomplished.

src/request.ts
```diff
@@ -1235,6 +1236,73 @@ class DatastoreRequest {
...
+    callback?: SaveCallback
+  ): void | Promise<CommitResponse> {
+    const transaction = this.datastore.transaction();
+    transaction.run(err => {
+      if (err) {
+        transaction.rollback();
+        callback!(err);
+        return;
+      }
+      arrify(entities)
```
stephenplusplus: Allowing multiple entities seems unnecessary. What do you think about just one per call to merge?

praveenqlogic01: @stephenplusplus Hi, I am following similar behavior from upsert and update function, it seems reasonable providing user to merge multiple entities.

stephenplusplus: I see, thanks!

src/request.ts
```diff
@@ -1235,6 +1236,73 @@ class DatastoreRequest {
...
+        .map(DatastoreRequest.prepareEntityObject_)
+        .forEach(
```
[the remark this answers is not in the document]
praveenqlogic01: @stephenplusplus Hi, refactor code with promise.all and used async flow.

bcoe:
@stephenplusplus have any feedback, has @praveenqlogic01 addressed your concerns?

stephenplusplus:
Yes, looks good to me. Thanks!

system-test/datastore.ts
```diff
@@ -305,6 +305,21 @@ describe('Datastore', () => {
...
+      const postKey = datastore.key(['Post', 1]);
+      await datastore.save({key: postKey, data: post});
+      const entity = {
+        key: postKey,
+        data: {
+          title: 'Updated',
+        },
+      };
+      await datastore.merge(entity);
+      const [entity2] = await datastore.get(postKey);
```
bcoe: it might be good to demonstrate in this system test that original values will be maintained if they're not included in `data`, e.g., that:

```
const obj = {title: 'original', status: 'neat};
const obj2 = {title: 'updated'}
merge(obj, obj2)
// will result inwill result in 
{title: 'updated', status; 'neat'}
```

praveenqlogic01: done!

BenWhitehead:
@bcoe It looks like your requested changes have been addressed, can you PTAL?

bcoe:
:shipit: this looks great to me.

Issue #105, opened 2018-06-11
How to generate URL-safe key?


bretthadley:
Hi, I can't find it in the documentation on how to generate a url-safe key?

Thanks

stephenplusplus:
Regarding the URL safe part, you can run any value through `encodeURIComponent`. I probably shouldn't comment on any best practices about generating the key-- @JustinBeckwith, should we ask someone from Datastore?

bretthadley:
Hi @stephenplusplus / @JustinBeckwith 

I need to generate the URL-safe key we see inside the datastore UI.

Regards, Brett.

callmehiphop:
@bretthadley sorry for the blast from the past, but did you ever figure out the answer to this?

jiren:
To generate URL - safe key required app engine protobuf Key structure (which include Reference and Path element). I am checking in detail how we can generate it.

AVaksman:
Closed with #456

PR #456, opened 2019-07-09
feat: add support for appengine legacy url encoding

### LINKED ISSUE #152: Query datastore entity by url-safe key or decode it to a searchable key
I'm trying to query datastore entity with NodeJS client API (https://www.npmjs.com/package/@google-cloud/datastore) given safe-url string provided by existing Go code. For some reason while other APIs (Go, Python,...) enable to decode/encode from/to safe-Url to key, NodeJS doesn't. Is there other way to do this? Thanks.

#### Environment details

  - OS:
  - Node.js version: 8.9.3
  - npm version: 6.2.0
  - `@google-cloud/datastore` version: 1.4.1

####

jiren:
Fixes #152 and #105 

- [x] Tests and linter pass


JustinBeckwith:
Thanks for the PR!  @crwilcox & @stephenplusplus could I trouble y'all to take a look here?

src/utils.ts
```diff
@@ -0,0 +1,164 @@
+import * as Protobuf from 'protobufjs';
```
stephenplusplus: Could these new classes be moved to `src/entity.ts`?

jiren: ok

jiren: @stephenplusplus  Moved into entity.ts

stephenplusplus:
@jiren looks like there is an issue with the samples test: https://source.cloud.google.com/results/invocations/f307d150-cef8-4d65-bb63-714d9931c311/targets/cloud-devrel%2Fclient-libraries%2Fnodejs%2Fpresubmit%2Fgoogleapis%2Fnodejs-datastore%2Fnode10%2Fsamples-test/log -- I'm not sure how this could be related to your changes, but other PRs don't seem to have the same failure. Any ideas?

jiren:
@stephenplusplus  looking into it.

Issue #460, opened 2019-07-25
Promise-returning methods within DatastoreRequest are incorrectly typed


ckknight:
#### Environment details

  - OS: macOS 10.14
  - Node.js version: 12.6.0
  - npm version: 6.9.0
  - `@google-cloud/datastore` version: 4.1.4
  - TypeScript version: 3.5.3

#### Steps to reproduce

```ts
const ds = new Datastore();
const value: Promise<any> = ds.delete(['x']);
```

Though `DatastoreRequest` does properly promisify its methods (https://github.com/googleapis/nodejs-datastore/commit/e57ed638640c28daa469f8bf546c2de62a6cc26e/src/request.ts#diff-31b74031f3c83e9e883fd930aaf87f54R1371), some of the types are inaccurate, such as `delete` with `keys` but no callback: https://github.com/googleapis/nodejs-datastore/commit/e57ed638640c28daa469f8bf546c2de62a6cc26e/src/request.ts#diff-31b74031f3c83e9e883fd930aaf87f54R312

PR #472, opened 2019-08-13
fix(types): fix and clean up types for request.ts

AVaksman:
Towards #460

- [x] Tests and linter pass

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -374,7 +374,7 @@ const urlSafeKey = new entity.URLSafeKey();
  * });
  */
 class Datastore extends DatastoreRequest {
-  clients_: Map<string, {}>;
+  clients_: Map<string, {[k: string]: Function}>;
   namespace?: string;
   projectId: string;
   defaultBaseUrl_: string;
--- a/src/index.ts
+++ b/src/index.ts
@@ -26,7 +26,7 @@
 
 import arrify = require('arrify');
 import {GoogleAuth, GoogleAuthOptions} from 'google-auth-library';
-import {GrpcClient} from 'google-gax';
+import {GrpcClient, ClientStub} from 'google-gax';
 import {ChannelCredentials} from '@grpc/grpc-js';
 import * as is from 'is';
 
@@ -374,7 +374,7 @@ const urlSafeKey = new entity.URLSafeKey();
  * });
  */
 class Datastore extends DatastoreRequest {
-  clients_: Map<string, {[k: string]: Function}>;
+  clients_: Map<string, ClientStub>;
   namespace?: string;
   projectId: string;
   defaultBaseUrl_: string;
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2018 Google LLC. All Rights Reserved.
+ * Copyright 2018 Google LLC.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/request.ts
+++ b/src/request.ts
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2014 Google Inc. All Rights Reserved.
+ * Copyright 2014 Google LLC.
  *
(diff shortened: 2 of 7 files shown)


src/request.ts
```diff
@@ -516,7 +523,7 @@ class DatastoreRequest {
   }
 
   insert(entities: Entities): Promise<CommitResponse>;
-  insert(entities: Entities, callback: CallOptions): void;
+  insert(entities: Entities, callback: CommitCallback): void;
```
callmehiphop: Should we follow suit with the other commit convenience methods and make a custom type for the callback/response here that extends the commit version?

src/request.ts
```diff
@@ -1330,45 +1325,53 @@ export type ProjectId = string | null | undefined;
...
-}
-export interface RunQueryStreamOptions {
-  gaxOptions?: CallOptions;
-  consistency?: 'strong' | 'eventual';
+  projectId?: ProjectId;
+  query?: QueryProto;
 }
-export interface SaveCallback {
-  (a?: Error | null, b?: Entity): void;
+export interface RunQueryStreamOptions extends RunQueryOptions {}
```
callmehiphop: Any reason why we wouldn't just re-use `RunQueryOptions`?

src/request.ts
```diff
@@ -1330,45 +1325,53 @@ export type ProjectId = string | null | undefined;
...
+export interface UpsertCallback extends CommitCallback {}
+export type UpsertResponse = CommitResponse;
 
 /*! Developer Documentation
  *
  * All async methods (except for streams) will return a Promise in the event
  * that a callback is omitted.
  */
-promisifyAll(DatastoreRequest);
+promisifyAll(DatastoreRequest, {exclude: ['request_']});
```
callmehiphop: Do we need this change? I thought that it excluded methods with trailing underscores by default?

AVaksman: You are right, forgot about it. 
Not needed!

src/request.ts
AVaksman: Just to keep the same "aesthetic" look as with `Save`/`Update`/`UpsertCallback` interfaces extend `CommitCallback` interface.
Not particular about it. Can change ether way?

src/request.ts
AVaksman: I think yes, nice catch!
Added `InsertCallback`/`Response` and `DeleteCallback`/`Response`

AVaksman:
Thank you for reviewing. Took care of the questions

src/index.ts
```diff
@@ -374,7 +374,7 @@ const urlSafeKey = new entity.URLSafeKey();
  * });
  */
 class Datastore extends DatastoreRequest {
-  clients_: Map<string, {}>;
+  clients_: Map<string, {[k: string]: Function}>;
```
callmehiphop: You could use `import {ClientStub} from 'google-gax';` here

src/request.ts
callmehiphop: No need, I was mostly curious!

src/request.ts
```diff
@@ -1330,38 +1323,50 @@ export type ProjectId = string | null | undefined;
 export interface RequestCallback {
   (
     a?: Error | null,
-    b?: AllocateIdsRequestResponse &
+    b?: google.datastore.v1.IAllocateIdsResponse &
```
callmehiphop: Hopefully I'm not opening a big can of worms, but should these response types be listed in this callback? Seems like maybe this should be a generic type? WDYT?

src/index.ts
AVaksman: Nice! Done
Thanks

src/request.ts
AVaksman: It should, it is internal anyways.

AVaksman: I did use `any` vs `Entity` which is masked `any` or vs extending from `GetCallback`.
I think it gives better code readability as "what you see is what you get" (personal preference).

callmehiphop: Sounds good to me!

PR #474, opened 2019-08-16
feat: support key re-use from serialized path

### LINKED ISSUE #59: Key paths are ambiguous, there is no trivial way to serialize a unique entity key!
Hey... So I've read a few issues on this. For example there's  ["Get Unique Entity Key String"](https://github.com/GoogleCloudPlatform/google-cloud-node/issues/625) about "urlsafe" keys. There the apparently accepted solution was to serialize and encode the key path...

However, key paths in nodejs are ambiguous! See [Key IDs Are Coming Back with String Values](https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2093).

#### Environment details

  - OS: Ubuntu Linux 16.04
  - Node.js version: 8.9.4
  - npm version: 5.6.0
  - @google-cloud/datastore version: 1.3.3

#### Steps to reproduce

```js
const key = db.key(["Post", "31337", "Comment", db.int("999999999999999999")]);
console.log(JSON.stringify(key));
// {"id":"999999999999999999","kind":"Comment","parent":{"name":"31337","kind":"Post","path":["Post","31337"]},"path":["Post","31337","Comment","999999999999999999"]}

const rekey = db.key(key.path);
console.log(JSON.stringify(rekey));
// {"name":"999999999999999999","kind":"Comment","parent":{"name":"31337","kind":"Post","path":["Post","31337"]},"path":["Post","31337","Comment","999999999999999999"]}
```

It's quite clear that any queries relying on the above serialized key path will fail!

I'm not sure if I can just JSON.parse() a key and use it in a query... If yes, that might be a solution, though this serialized format is ridiculously verbose to use as a foreign key or even as a transmission format.

I can work around this by NOT using entity groups AT ALL (cutting out one of the ways I could optimize a Datastore db), and only having references to fixed Kinds (coming from a relational background I can live with this one)... In this case, I can always just store or send a numeric Id (as a decimal string due to JS number limitations). But still, it's kinda painful compared to having a globally unique serializable Id I could easily use for caching, references, etc...

Thanks!

stephenplusplus:
Fixes #59 

This exposes a `serialized` property on a Key object. It will correctly differentiate between a key name and id, so that you can re-create a key with it:

```js
const key = datastore.key({
  namespace: 'My-NS',
  path: ['Company', 123]
);
// Later...
const key = datastore.key(key.serialized);
```

### DIFF
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -223,8 +223,7 @@ export namespace entity {
    *   path: ['Company', 123]
    * });
    *
-   * @example
-   * <caption>Serialize the key for later re-use.</caption>
+   * @example <caption>Serialize the key for later re-use.</caption>
    * const {Datastore} = require('@google-cloud/datastore');
    * const datastore = new Datastore();
    * const key = datastore.key({
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -300,7 +300,9 @@ export namespace entity {
           };
 
           if (this.parent) {
-            serializedKey.path = this.parent.serialized.path.concat(serializedKey.path);
+            serializedKey.path = this.parent.serialized.path.concat(
+              serializedKey.path
+            );
           }
 
           return serializedKey;
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -222,6 +222,19 @@ export namespace entity {
    *   namespace: 'My-NS',
    *   path: ['Company', 123]
    * });
+   *
+   * @example
+   * //-
+   * // Serialize the key for later re-use.
+   * //-
+   * const {Datastore} = require('@google-cloud/datastore');
+   * const datastore = new Datastore();
+   * const key = datastore.key({
+   *   namespace: 'My-NS',
+   *   path: ['Company', 123]
+   * });
+   * // Later...
+   * const key = datastore.key(key.serialized);
    */
   export class Key {
     namespace?: string;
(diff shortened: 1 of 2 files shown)


src/entity.ts
```diff
@@ -271,6 +285,29 @@ export namespace entity {
...
         },
       });
+
+      // `serialized` is computed on demand to consider any changes that may
+      // have been made to the key.
+      /**
+       * @name Key#serialized
+       * @type {array}
+       */
+      Object.defineProperty(this, 'serialized', {
```
JustinBeckwith: Why use `Object.defineProperty` as opposed to a regular getter?

src/entity.ts
```diff
@@ -222,6 +222,19 @@ export namespace entity {
    *   namespace: 'My-NS',
    *   path: ['Company', 123]
    * });
+   *
+   * @example
+   * //-
+   * // Serialize the key for later re-use.
```
callmehiphop: I think our new docs site supports `<caption>` for this

src/entity.ts
stephenplusplus: Give it a few years-- `Object.defineProperty` will be in style again, and the kids will be wearing it on their hoodies.

src/entity.ts
```diff
@@ -222,6 +222,17 @@ export namespace entity {
    *   namespace: 'My-NS',
    *   path: ['Company', 123]
    * });
+   *
+   * @example
+   * <caption>Serialize the key for later re-use.</caption>
```
callmehiphop: I'm not sure if this would impact the formatting, but usually caption is listed on the same line as the example tag. Might want to generate this locally and check

PR #480, opened 2019-08-27
feat: support DSInt / DSDouble recognition from JSON

### LINKED ISSUE #59: Key paths are ambiguous, there is no trivial way to serialize a unique entity key!

hermanbanken:
We had issues comparable to #59 in the past. IMO the proposed solution in #474 is only half of the solution: the missing part is more forgiving deserialisation. Note also that users like @beaulac [have reported](https://github.com/googleapis/nodejs-datastore/issues/59#issuecomment-366739401) creating work-arounds for this specifically. This PR solves this by extending the constructor of Key.

Fixes #59. Serialization comes with deserialization, and the current
deserialization did not deal well with plain JavaScript objects like
those returned from JSON deserialization.

Given this implementation is it easy & safe to transfers Keys over
networks and other non-JavaScript interfaces.

- [x] Tests and linter pass

### DIFF
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -63,8 +63,14 @@ export namespace entity {
    * const aDouble = datastore.double(7.3);
    */
   export class Double {
+    type: string;
     value: number;
     constructor(value: number) {
+      /**
+       * @name Double#type
+       * @type {string}
+       */
+      this.type = 'DatastoreDouble';
       /**
        * @name Double#value
        * @type {number}
@@ -85,6 +91,23 @@ export namespace entity {
   }
 
   /**
+   * Check if a value is a Datastore Double object converted from JSON.
+   *
+   * @private
+   * @param {*} value
+   * @returns {boolean}
+   */
+  export function isDsDoubleLike(value: unknown) {
+    const maybeDsDouble = value as Double;
+    return (
+      isDsDouble(maybeDsDouble) ||
+      (is.object(maybeDsDouble) &&
+        is.number(maybeDsDouble.value) &&
+        maybeDsDouble.type === 'DatastoreDouble')
+    );
+  }
+
+  /**
    * Build a Datastore Int object. For long integers, a string can be provided.
    *
    * @class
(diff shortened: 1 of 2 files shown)


hermanbanken:
That would be helpful if you transfer entities themselves, but not needed in the case of Keys: or does Google support Keys with Double id's? 😨 

It might be a nice addition for other uses, but I don't want to complicate this PR with that addition.

hermanbanken:
Sure, feel free to add onto this.

I was not considering such API misusage 😇 (`new entity.Key(["User", new Double("3.141592")])`)

stephenplusplus:
@hermanbanken would you please take a look at my changes? According to @googlebot, we'd also need you to confirm you're okay with them as well 👍

hermanbanken:
@googlebot I consent.

stephenplusplus:
Thank you @hermanbanken! Sorry it took me a while to get back to this.

PR #488, opened 2019-09-05
feat!: throw error with out of bounds integer values instead of truncating, add custom 'integerValue' type cast option

### LINKED ISSUE #147: Inconsistent processing of large numbers from other languages.
Because javascript cannot process large numbers, it becomes extremely difficult to work with integrated systems.

We have Java applications storing large `long` values to Datastore, which now has a type `Integer` in Datastore.  These are then unable to be read by systems using node.

While the node version can also STORE large values by using the `datastore.int()` method, when reading the value out of the database, it is automatically read as a `number` javascript type, and therefore truncated.

AVaksman:
Fixes #147

- [x] Tests and linter pass

As per discussion, implementation follows design in the [document](https://docs.google.com/document/d/1xyvXqM4fujftccCpS5NGQEm6Gci46L4beZshw183RIk/edit#heading=h.81g19iglxc7s)

- On default detect if value is out of bounds of `Number`
- Allow user to pass a custom `typeCastFunction` to be used to convert integer values
    - Give user an option to specify property names to be converted by `typeCastFunction`


bcoe:
@crwilcox were we comfortable with making this a major release of datastore? I know we're trying to keep major bumps infrequent, but this seems like a reasonably useful refactor to consider one.

src/entity.ts
```diff
@@ -326,11 +326,67 @@ export namespace entity {
...
+   *     names to be converted using `typeCastFunction`.
+   */
+  function decodeIntegerValue(
+    valueProto: ValueProto,
+    typeCastConfig?: TypeCastConfig
+  ) {
+    let customCast = false;
+    if (typeCastConfig) {
+      if (
+        !typeCastConfig.typeCastFunction ||
```
callmehiphop: Just a nit, but you could just do the `typeof` check here.

AVaksman: True, thanks
Done

AVaksman:
> So just a general note, I didn't realize that the type cast function would only apply to integers. Not sure if I missed a separate conversation leading to this PR, but I think this would be useful for all data types, not just integers.

My logic here is, that library doesn't have an issue with correctly obtaining all other value types and if any conversions/manipulations are required, the user can do them after getting the values from this library.

With regards to numeric types there is an issue of obtaining accurate large integer value.

Another point is, I wouldn't want to incorporate third party code more than I have to. Not to subjugate the library to performance impact, bugs, etc... from third party.

callmehiphop:
@AVaksman makes sense, I might be bikeshedding but if this only applies to integers we might want to revisit the name? Though if no one else feels the same way I don't feel super strongly about it 😄

AVaksman:
Refactored to 
`typeCastConfig` => `integerTypeCastOptions` 
`typeCastFunction` => `integerTypeCastFunction`

AVaksman:
> Based on the PR description, I can't really figure out what the breaking change is. Can you add a really clear message in there for the release notes, including before code and after code?

The breaking part is that now in case of `integerValue` is out of bounds of `Number` the API will throw an error instead of truncating the value as before.

Users who needed to accurately obtain out of bounds integer values, couldn't. So they probably didn't use this functionality, hence practically speaking this is not breaking change.

src/entity.ts
```diff
@@ -372,7 +435,7 @@ export namespace entity {
       }
 
       case 'integerValue': {
-        return Number(value);
+        return decodeIntegerValue(valueProto, integerTypeCastOptions);
```
stephenplusplus: This is definitely a more robust option, but was just returning the DS.Int type considered?

stephenplusplus: Sorry, I mixed up Datastore's "Int" with Spanner's: https://github.com/googleapis/nodejs-spanner/blob/f8ad0ece28e99dd9c6d7e985de56bd1b376f4712/src/codec.ts#L139 -- for simplicity's sake, would it be worth considering using that format instead? When a value is out of bounds, it will throw. The value is accessible via the "value" property in string form.

JustinBeckwith:
@AVaksman ah in that case - can you drop the `!` from the commit title and PR title?

src/entity.ts
AVaksman: * I do like the option of ability to still access the value on default (without extra call) in case of value is out of balance.
* it will definitely be a breaking change
* it will add complexity to the user's code
   will need to `try=>catch` for `valueOf` and if `error` still pass to custom `typeCastFunction`.
* Is it overkill to combine these two approaches
     * return Spanner's style object for `integerValue`.
     * for `valueOf` on default throw in case of out of bounds.
     * able to pass integerTypeCastFunction and `valueOf` will return the custom function result.

AVaksman: WDYT?

AVaksman:
Roger that

src/entity.ts
```diff
@@ -327,11 +327,68 @@ export namespace entity {
...
 
+  /**
+   * Convert a protobuf `integerValue`.
+   *
+   * @private
+   * @param {object} valueProto The protobuf `integerValue` to convert.
+   * @param {object} integerTypeCastOptions Config for custom `integerValue` cast.
+   * @property {function} {integerTypeCastOptions.integerTypeCastFunction} A custom user
+   *     provided function to convert `integerValue`.
+   * @property {sting|string[]} {typeCastConig.names} `Entity` property
```
jkwlui: `integerTypeCastOptions.names`

src/entity.ts
```diff
@@ -327,11 +327,68 @@ export namespace entity {
     return value instanceof entity.Key;
   }
 
+  /**
+   * Convert a protobuf `integerValue`.
+   *
+   * @private
+   * @param {object} valueProto The protobuf `integerValue` to convert.
+   * @param {object} integerTypeCastOptions Config for custom `integerValue` cast.
+   * @property {function} {integerTypeCastOptions.integerTypeCastFunction} A custom user
```
jkwlui: `@param {function} [integerTypeCastOptions.integerTypeCastFunction] ...`

src/entity.ts
```diff
@@ -327,11 +327,68 @@ export namespace entity {
...
+    }
+  }
+
   /**
    * Convert a protobuf Value message to its native value.
    *
    * @private
    * @param {object} valueProto The protobuf Value message to convert.
+   * @param {object} integerTypeCastOptions Config for custom `integerValue` cast.
+   * @property {function} {integerTypeCastOptions.integerTypeCastFunction} A custom user
```
jkwlui: Same as above. This should be `@param`

src/entity.ts
```diff
@@ -719,6 +794,11 @@ export namespace entity {
    * @param {object[]} results The response array.
    * @param {object} results.entity An entity object.
    * @param {object} results.entity.key The entity's key.
+   * @param {object} integerTypeCastOptions Config for custom `integerValue` cast.
+   * @property {function} {integerTypeCastOptions.integerTypeCastFunction} A custom user
```
jkwlui: Same as above

AVaksman: thanks, done

src/entity.ts
AVaksman: thanks, done

src/entity.ts
AVaksman: thanks, done

src/entity.ts
AVaksman: thanks, done

src/entity.ts
stephenplusplus: Did we ever make progress on this, maybe in another chat?

AVaksman: Quick summary
Would like to hear the thoughts on the approach
1. Throw on default, let user pass integerTypeCastFunction (currently implemented in this PR)
2. Return an object `Integer` where 
    * `#valueOf` would try to convert to `Number` and throw if out of bounds
    * `#value` returns `string` representation of the value
3. Combine tow approaches above 
    * do return an `Integer` object 
    * let user provide integerTypeCast function to be used by `#valueOf`
    * still return string via `#value`

stephenplusplus: My vote would go for #&shy;3. Ping to @callmehiphop and @bcoe for more votes.

BenWhitehead:
@jkwlui Does everything look good here for you after the latest fixes?

AVaksman:
Quick summary
Would like to hear the thoughts on the approach
1. Throw on default, let user pass integerTypeCastFunction (currently implemented in this PR)
2. Return an object `Integer` similar to [Spanner](https://github.com/googleapis/nodejs-spanner/blob/f8ad0ece28e99dd9c6d7e985de56bd1b376f4712/src/codec.ts#L139), where 
    * `#valueOf` would try to convert to `Number` and throw if out of bounds
    * `#value` returns `string` representation of the value
3. Combine tow approaches above 
    * do return an `Integer` object 
    * let user provide integerTypeCast function to be used by `#valueOf`
    * still return string via `#value`

from @stephenplusplus 
> My vote would go for #­3.

stephenplusplus:
@AVaksman maybe it'll just be us deciding 😆 What is your preference?

src/entity.ts
bcoe: My vote is starting simple with option 1, as a breaking change, and we can move towards an object representation in the future, if there's good reason.

AVaksman:
@stephenplusplus I do like option 3 as well, @bcoe 's [vote](https://github.com/googleapis/nodejs-datastore/pull/488#discussion_r331174097) goes to # 1

JustinBeckwith:
@AVaksman @stephenplusplus @bcoe what's the next steps here?

AVaksman:
The next step would be to decide between approach in this PR vs #516

stephenplusplus:
I don't see us going back to this after where we landed in #516, so I'll just close for now to avoid any confusion.

PR #492, opened 2019-09-11
fix: emit encoding errors on user stream

### LINKED ISSUE #491: Uncatchable error is thrown if undefined is given on query filter (JS built version)
#### Environment details

  - OS: macOS Mojave
  - Node.js version: v10.16.0
  - npm version: 6.9.0
  - `@google-cloud/datastore` version: 4.2.0
  - Built version

#### Steps to reproduce

If we give `undefined` to a query filter, an uncatchable error is thrown: `Error: Unsupported field value, undefined, was provided.`, and program crashes.

Problem occurs only in built version, because Typescript doesn't compile with this scenario:

```js
let query = new FakeQuery();
query.namespace = 'namespace';
query = extend(true, new FakeQuery(), query);
query.filter('myfield', undefined); // => error TS2345: Argument of type 'undefined' is not assignable to parameter of type '{}'.
```

To be able to catch the error, the library instead of doing:

```js
// src/request/ts: l.788
stream.once('reading', () => {
  makeRequest(query);
});
```

could do this:

```js
// src/request/ts: l.788
stream.once('reading', () => {
  try {
    makeRequest(query);
  } catch (err) {
    stream.emit('error', err);
  }
});
```

For Typescript it doesn't change anything. But for the JS built version, the error becomes catchable.

I can send a pull request if you want.

callmehiphop:
Fixes #491 

- [x] Tests and linter pass

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -715,6 +715,8 @@ class DatastoreRequest {
       try {
         reqOpts.query = entity.queryToQueryProto(query);
       } catch (e) {
+        // using setImmediate here to make sure this doesn't throw a
+        // synchronous error
         setImmediate(onResultSet, e);
         return;
       }
--- a/src/request.ts
+++ b/src/request.ts
@@ -710,9 +710,14 @@ class DatastoreRequest {
     query = extend(true, new Query(), query);
 
     const makeRequest = (query: Query) => {
-      const reqOpts: RequestOptions = {
-        query: entity.queryToQueryProto(query),
-      };
+      const reqOpts = {} as RequestOptions;
+
+      try {
+        reqOpts.query = entity.queryToQueryProto(query);
+      } catch (e) {
+        setImmediate(onResultSet, e);
+        return;
+      }
 
       if (options.consistency) {
         const code = CONSISTENCY_PROTO_CODE[options.consistency.toLowerCase()];
--- a/test/request.ts
+++ b/test/request.ts
@@ -819,6 +819,18 @@ describe('Request', () => {
           })
           .emit('reading');
       });
+
+      it('should emit an error when encoding fails', done => {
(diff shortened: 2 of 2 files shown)


src/request.ts
```diff
@@ -710,9 +710,14 @@ class DatastoreRequest {
...
     const makeRequest = (query: Query) => {
-      const reqOpts: RequestOptions = {
-        query: entity.queryToQueryProto(query),
-      };
+      const reqOpts = {} as RequestOptions;
+
+      try {
+        reqOpts.query = entity.queryToQueryProto(query);
+      } catch (e) {
+        setImmediate(onResultSet, e);
```
JustinBeckwith: Why the `setImmediate`? could I bother you for a comment explaining so the next soul doesn't trip?

callmehiphop: Personal preference really. In my experience I've found unit testing callbacks that can be called early (and synchronously) to sometimes give weird results. I'm not married to it if you would prefer to go without it.

bcoe: If I understand correctly, this is addressing edge-cases discussed in [this post about zalgo](https://blog.izs.me/2013/08/designing-apis-for-asynchrony)?

Definitely a bit icky, as the language evolves, I think I'll be moving towards code more like this:

```
try {
  reqOpts.query = await entityToQueryProto(query); 
} catch (e) {
   onResultSet(e);
  return;
}
```

If a method could have an async error, treat it as an async method in general .... _but_, I can understand not wanting to take on this refactor.

I'd just suggest adding a comment, speaking to zalgo, e.g., _this ensures that the API doesn't sometime throw a synchronous error._

callmehiphop: Done! Thanks for the link to that post, pretty nice read!

JustinBeckwith (approved):

PR #496, opened 2019-09-12
fix!: keyToLegacyUrlsafe is now an async method

### LINKED ISSUE #473: keyToLegacyUrlsafe running on Cloud Functions uses wrong projectId?
#### Environment details

  - OS: Google Cloud Functions
  - Node.js version: 8.16.0
  - npm version: 6.10.3
  - `@google-cloud/datastore` version: 4.2.0

#### Steps to reproduce

I'm using keyToLegacyUrlsafe to convert keys to the legacy App Engine format so I can interact with my legacy Python 2.7 App Engine code via API. 

I am calling this from a Cloud Function, triggered by Pub/Sub. 

My App Engine code receives the urlsafe string but complains that the project is not in the correct format - the error states `Application Id (app) format is invalid: '{{projectId}}'`. The key's path is correct.

I have worked around the problem in my Python App Engine code for now, but I believe the root cause of this is in either the node module or Cloud Functions itself. At no point do I set the project ID myself in the code, it's picked up from the runtime environment.

Let me know if you need more details, I'll be investigating further on my side today and I'll update this if I find anything relevant.

laljikanjareeya:
Fixes #473 

- [x] Tests and linter pass

While calling `keyToLegacyUrlsafe` function it's not binding `{{projectId}}` to actual one as it does not have any external/api call.
Added `auth.getProjectId` to get projectId and change it to callback and promise support as `auth.getProjectId` function's behavior is same.

**BREAKING CHANGE:** 

- keyToLegacyUrlsafe is rename to keyToLegacyUrlSafe

- keyToLegacyUrlSafe is now an async method


test/index.ts
```diff
@@ -630,10 +630,46 @@ describe('Datastore', () => {
...
+      // tslint:disable-next-line: no-any
+      (datastore.auth as any).getProjectId = (callback: Function) => {
+        callback(null, 'project-id');
+      };
+      datastore.keyToLegacyUrlsafe(key, locationPrefix, (err, urlSafekey) => {
+        assert.strictEqual(urlSafekey, base64EndocdedUrlsafeKey);
+      });
+    });
+
+    it('should take projectId from auth if not exists', () => {
```
AVaksman: Based on the implementation of the function, I do not think this test is needed.

AVaksman:
Besides the extra test
LGTM

src/index.ts
```diff
@@ -870,6 +918,27 @@ class Datastore extends DatastoreRequest {
   Transaction = Transaction;
 }
 
+/*! Developer Documentation
+ *
+ * All async methods (except for streams) will return a Promise in the event
+ * that a callback is omitted.
+ */
+promisifyAll(Datastore, {
```
bcoe: could we use `util.promisify`?

laljikanjareeya: > could we use `util.promisify`?

@bcoe  for consistency `promisifyAll` is used as currently in all the library we are using the same.

bcoe: :+1: works for me, might be worth doing some work on the `util` module to make sure `promisfyAll` takes advantage of `util.promisify`.

test/index.ts
laljikanjareeya: removed!!

src/index.ts
```diff
@@ -744,26 +755,63 @@ class Datastore extends DatastoreRequest {
...
+    const locationPrefix =
+      typeof locationPrefixOrCallback === 'string'
+        ? locationPrefixOrCallback
+        : '';
+    callback =
+      typeof locationPrefixOrCallback === 'function'
+        ? locationPrefixOrCallback
+        : callback;
+    this.auth.getProjectId((err, projectId) => {
+      callback!(err, urlSafeKey.legacyEncode(projectId!, key, locationPrefix));
```
stephenplusplus: When there's an error, we should not return the best-effort encoded key. That could lead users to believe it worked successfully.

AVaksman: @laljikanjareeya 
something like
```ts
this.auth.getProjectId((err, projectId) => {
  if (err) {
    callback!(err);
    return;
  }
  callback!(null, urlSafeKey.legacyEncode(projectId!, key, locationPrefix));
});
```
and add a test, something along
```ts
    it('should not return urlSafeKey to user if auth.getProjectId erros', () => {
      const error = new Error('Some error');
      // tslint:disable-next-line: no-any
      (datastore.auth as any).getProjectId = (callback: Function) => {
        callback(error);
      };
      datastore.keyToLegacyUrlsafe({} as entity.Key, (err, urlSafekey) => {
        assert.strictEqual(err, error);
        assert.strictEqual(urlSafekey, undefined);
      });
    });

```

BenWhitehead:
@stephenplusplus Can you please re-review this after the recent updates?

bcoe:
This looks good to me, we should call out what is specifically breaking in the message body of what is landed:

```
BREAKING CHANGE: keyToLegacyUrlsafe is now an async method
```

src/index.ts
```diff
@@ -932,3 +1005,7 @@ export interface DatastoreOptions extends GoogleAuthOptions {
   apiEndpoint?: string;
   sslCreds?: ChannelCredentials;
 }
+
+export interface KeyToLegacyUrlsafeCallback {
+  (err?: Error | null, urlSafekey?: string): void;
```
stephenplusplus: I'm not sure if there's an objective answer to this, but to me, `urlSafe` feels more correct than `urlsafe`. This and the previous line show two different interpretations of when to camelcase:

```
KeyToLegacyUrlsafeCallback
urlSafekey
```

I would expect:

```diff
- KeyToLegacyUrlsafeCallback
+ KeyToLegacyUrlSafeCallback

- urlSafekey
+ urlSafeKey
```

test/index.ts
```diff
@@ -630,10 +630,42 @@ describe('Datastore', () => {
...
+      });
+      const locationPrefix = 's~';
+      const base64EndocdedUrlsafeKey =
+        'agxzfnByb2plY3QtaWRyDgsSBFRhc2siBFRlc3QM';
+      // tslint:disable-next-line: no-any
+      (datastore.auth as any).getProjectId = (callback: Function) => {
+        callback(null, 'project-id');
+      };
+      datastore.keyToLegacyUrlsafe(key, locationPrefix, (err, urlSafekey) => {
+        assert.strictEqual(urlSafekey, base64EndocdedUrlsafeKey);
```
stephenplusplus: Please add an `assert.ifError(err)` above this line to catch unexpected errors.

test/index.ts
```diff
@@ -630,10 +630,42 @@ describe('Datastore', () => {
...
-        datastore.keyToLegacyUrlsafe(key),
-        'agpwcm9qZWN0LWlkcg4LEgRUYXNrIgRUZXN0DA'
-      );
+      const base64EndocdedUrlsafeKey = 'agpwcm9qZWN0LWlkcg4LEgRUYXNrIgRUZXN0DA';
+      // tslint:disable-next-line: no-any
+      (datastore.auth as any).getProjectId = (callback: Function) => {
+        callback(null, 'project-id');
+      };
+      datastore.keyToLegacyUrlsafe(key, (err, urlSafekey) => {
+        assert.strictEqual(urlSafekey, base64EndocdedUrlsafeKey);
```
stephenplusplus: Please add an `assert.ifError(err)` above this line to catch unexpected errors.

src/index.ts
laljikanjareeya: Done!

src/index.ts
```diff
@@ -730,6 +731,16 @@ class Datastore extends DatastoreRequest {
     return Datastore.isKey(value);
   }
 
+  keyToLegacyUrlsafe(key: entity.Key, locationPrefix?: string): Promise<string>;
```
stephenplusplus: In these other places, I believe we need to now update to the new camelcase system, `UrlSafe`/`urlSafe`

laljikanjareeya: Rename the method name, also updated to description.

src/index.ts
```diff
@@ -744,26 +755,67 @@ class Datastore extends DatastoreRequest {
...
+   * datastore.keyToLegacyUrlSafe(key, locationPrefix, (err, urlSafeKey) => {
+   *   if (err) {
+   *     // Error handling omitted.
+   *   }
+   *   console.log(urlSafeKey);
+   *
+   * //-
+   * // If the callback is omitted, we'll return a Promise.
+   * //-
+   * const {Datastore} = require('@google-cloud/datastore');
```
stephenplusplus: These boilerplate lines can be skipped as well.

src/index.ts
```diff
@@ -744,26 +755,67 @@ class Datastore extends DatastoreRequest {
...
+   * const datastore = new Datastore();
+   * const key = datastore.key(['Task', 123]);
+   * const locationPrefix = 's~';
+   *
+   * const urlSafeKey = await datastore.keyToLegacyUrlSafe(key, locationPrefix);
```
stephenplusplus: To be consistent with our other examples in the repo, let's show this as a promise:

```js
datastore.keyToLegacyUrlSafe(key).then((data) => {
  const urlSafeKey = data[0];
});
```

src/index.ts
```diff
@@ -744,26 +755,67 @@ class Datastore extends DatastoreRequest {
...
+   *   }
+   *   console.log(urlSafeKey);
+   *
+   * //-
+   * // If the callback is omitted, we'll return a Promise.
+   * //-
+   * const {Datastore} = require('@google-cloud/datastore');
+   * const datastore = new Datastore();
+   * const key = datastore.key(['Task', 123]);
+   * const locationPrefix = 's~';
```
stephenplusplus: I would skip showing the locationPrefix for this promise example.

src/index.ts
```diff
@@ -744,26 +755,62 @@ class Datastore extends DatastoreRequest {
    *  The location prefix of an App Engine project ID.
    *  Often this value is 's~', but may also be 'e~', or other location prefixes
    *  currently unknown.
-   * @returns {string} base64 endocded urlsafe key.
+   * @param {function} callback The callback function.
+   * @param {?error} callback.err An error returned while making this request
+   * @param {string} callback.urlSafeKey base64 endocded urlsafe key.
```
stephenplusplus: Sorry :(, I just noticed this one.

```diff
- base64 endocded urlsafe key.
+ A Base64-encoded URL-safe key.
```

bcoe:
looks like this is ready to land, but let's hold off on landing for a tiny bit, and see if we can batch it up with some breaking changes @AVaksman is working on.

stephenplusplus:
@bcoe should we still hold off? Feels iffy to keep this floating around 🤷‍♂

PR #516, opened 2019-10-13
feat!: throw error with out of bounds integer values, optionally wrap into DsInt or provide a custom 'integerValue' type cast options

### LINKED ISSUE #147: Inconsistent processing of large numbers from other languages.

AVaksman:
Fixes #147

- [x] Tests and linter pass

`integerValue` is now decoded into `datastore.Int` object

`entity.Int#valueof()` now
- On default will try to convert to `Number`
    - will detect if value is out of bounds of `Number`
- Allow user to pass a custom `typeCastFunction` to be used to convert integer values
    - Give user an option to specify property names to be converted by `typeCastFunction`

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
     "intelli-espower-loader": "^1.0.1",
     "jsdoc": "^3.6.2",
     "jsdoc-fresh": "^1.0.1",
+    "jsdoc-region-tag": "^1.0.2",
     "linkinator": "^1.5.0",
     "mocha": "^6.1.4",
     "mv": "^2.1.1",
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
   "dependencies": {
     "@google-cloud/projectify": "^1.0.0",
     "@google-cloud/promisify": "^1.0.0",
-    "@grpc/grpc-js": "0.6.9",
+    "@grpc/grpc-js": "0.6.10",
     "@types/duplexify": "^3.6.0",
     "@types/long": "^4.0.0",
     "arrify": "^2.0.1",
--- a/package.json
+++ b/package.json
@@ -58,7 +58,8 @@
     "google-gax": "^1.7.5",
     "is": "^3.3.0",
     "split-array-stream": "^2.0.0",
-    "stream-events": "^1.0.5"
+    "stream-events": "^1.0.5",
+    "typescript": "~3.7.0"
   },
   "devDependencies": {
     "@types/extend": "^3.0.1",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@google-cloud/datastore",
   "description": "Cloud Datastore Client Library for Node.js",
-  "version": "4.4.1",
+  "version": "4.5.0",
   "license": "Apache-2.0",
   "author": "Google LLC.",
   "engines": {
@@ -48,7 +48,7 @@
   "dependencies": {
     "@google-cloud/projectify": "^1.0.0",
     "@google-cloud/promisify": "^1.0.0",
-    "@grpc/grpc-js": "^0.6.0",
+    "@grpc/grpc-js": "0.6.9",
     "@types/duplexify": "^3.6.0",
     "@types/long": "^4.0.0",
     "arrify": "^2.0.1",
--- a/package.json
+++ b/package.json
@@ -58,8 +58,7 @@
     "google-gax": "^1.7.5",
     "is": "^3.3.0",
     "split-array-stream": "^2.0.0",
-    "stream-events": "^1.0.5",
-    "typescript": "~3.7.0"
+    "stream-events": "^1.0.5"
   },
   "devDependencies": {
     "@types/extend": "^3.0.1",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@google-cloud/datastore",
   "description": "Cloud Datastore Client Library for Node.js",
-  "version": "4.5.0",
+  "version": "4.5.2",
   "license": "Apache-2.0",
   "author": "Google LLC.",
   "engines": {
@@ -48,7 +48,7 @@
   "dependencies": {
     "@google-cloud/projectify": "^1.0.0",
     "@google-cloud/promisify": "^1.0.0",
-    "@grpc/grpc-js": "0.6.10",
+    "@grpc/grpc-js": "0.6.11",
     "@types/duplexify": "^3.6.0",
     "@types/long": "^4.0.0",
     "arrify": "^2.0.1",
@@ -90,6 +90,6 @@
     "proxyquire": "^2.1.0",
     "sinon": "^7.3.2",
     "tmp": "0.1.0",
-    "typescript": "~3.6.0"
+    "typescript": "~3.7.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@google-cloud/datastore",
   "description": "Cloud Datastore Client Library for Node.js",
-  "version": "4.4.0",
+  "version": "4.4.1",
   "license": "Apache-2.0",
   "author": "Google LLC.",
   "engines": {
@@ -55,7 +55,7 @@
     "concat-stream": "^2.0.0",
     "extend": "^3.0.2",
     "google-auth-library": "^5.0.0",
-    "google-gax": "^1.6.3",
+    "google-gax": "^1.7.5",
     "is": "^3.3.0",
     "split-array-stream": "^2.0.0",
     "stream-events": "^1.0.5"
--- a/samples/concepts.js
+++ b/samples/concepts.js
@@ -1025,6 +1025,10 @@ class Transaction extends TestHelper {
         accounts[0].balance.valueOf(),
         originalBalance - amountToTransfer
       );
+      assert.equal(
+        accounts[1].balance,
+        originalBalance + amountToTransfer
+      );
       assert.strictEqual(
         accounts[1].balance.valueOf(),
         originalBalance + amountToTransfer
--- a/samples/concepts.js
+++ b/samples/concepts.js
@@ -731,8 +731,8 @@ class Query extends TestHelper {
       const percentCompletes = [];
       const [tasks] = await datastore.runQuery(query);
       tasks.forEach(task => {
-        priorities.push(task.priority.valueOf());
-        percentCompletes.push(task.percent_complete.valueOf());
+        priorities.push(task.priority);
+        percentCompletes.push(task.percent_complete);
       });
 
       return {
@@ -1017,16 +1017,12 @@ class Transaction extends TestHelper {
       const accounts = results.map(result => result[0]);
       // Restore `datastore` to the mock API.
       datastore = datastoreMock;
-      // eslint-disable-next-line node/no-deprecated-api
-      assert.equal(accounts[0].balance, originalBalance - amountToTransfer);
       assert.strictEqual(
-        accounts[0].balance.valueOf(),
+        accounts[0].balance,
         originalBalance - amountToTransfer
       );
-      // eslint-disable-next-line node/no-deprecated-api
-      assert.equal(accounts[1].balance, originalBalance + amountToTransfer);
       assert.strictEqual(
-        accounts[1].balance.valueOf(),
+        accounts[1].balance,
         originalBalance + amountToTransfer
       );
     } catch (err) {
--- a/samples/concepts.js
+++ b/samples/concepts.js
@@ -731,8 +731,8 @@ class Query extends TestHelper {
       const percentCompletes = [];
       const [tasks] = await datastore.runQuery(query);
       tasks.forEach(task => {
-        priorities.push(task.priority);
-        percentCompletes.push(task.percent_complete);
+        priorities.push(task.priority.valueOf());
+        percentCompletes.push(task.percent_complete.valueOf());
       });
 
       return {
@@ -1018,11 +1018,11 @@ class Transaction extends TestHelper {
       // Restore `datastore` to the mock API.
       datastore = datastoreMock;
       assert.strictEqual(
-        accounts[0].balance,
+        accounts[0].balance.valueOf(),
         originalBalance - amountToTransfer
       );
       assert.strictEqual(
-        accounts[1].balance,
+        accounts[1].balance.valueOf(),
         originalBalance + amountToTransfer
       );
     } catch (err) {
--- a/samples/concepts.js
+++ b/samples/concepts.js
@@ -1017,6 +1017,10 @@ class Transaction extends TestHelper {
       const accounts = results.map(result => result[0]);
       // Restore `datastore` to the mock API.
       datastore = datastoreMock;
+      assert.equal(
+        accounts[0].balance,
+        originalBalance - amountToTransfer
+      );
       assert.strictEqual(
         accounts[0].balance.valueOf(),
         originalBalance - amountToTransfer
--- a/samples/concepts.js
+++ b/samples/concepts.js
@@ -1017,18 +1017,14 @@ class Transaction extends TestHelper {
       const accounts = results.map(result => result[0]);
       // Restore `datastore` to the mock API.
       datastore = datastoreMock;
-      assert.equal(
-        accounts[0].balance,
-        originalBalance - amountToTransfer
-      );
+      // eslint-disable-next-line node/no-deprecated-api
+      assert.equal(accounts[0].balance, originalBalance - amountToTransfer);
       assert.strictEqual(
         accounts[0].balance.valueOf(),
         originalBalance - amountToTransfer
       );
-      assert.equal(
-        accounts[1].balance,
-        originalBalance + amountToTransfer
-      );
+      // eslint-disable-next-line node/no-deprecated-api
+      assert.equal(accounts[1].balance, originalBalance + amountToTransfer);
       assert.strictEqual(
         accounts[1].balance.valueOf(),
         originalBalance + amountToTransfer
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -111,10 +111,10 @@ export namespace entity {
    *
    * @class
    * @param {number|string} value The integer value.
-   * @param {object} typeCastOptions Config for custom `integerValue` cast.
-   * @param {boolean} [wrapNumbers=false] Indicates if the numbers should be
-   *     wrapped in Int wrapper.
-   * @param {function} [typeCastOptions.integerTypeCastFunction] A custom user
+   * @param {object} [typeCastOptions] Configuration to convert
+   *     values of `integerValue` type to a custom value. Must provide an
+   *     `integerTypeCastFunction` to handle `integerValue` conversion.
+   * @param {function} typeCastOptions.integerTypeCastFunction A custom user
    *     provided function to convert `integerValue`.
    * @param {sting|string[]} [typeCastOptions.properties] `Entity` property
    *     names to be converted using `integerTypeCastFunction`.
@@ -151,15 +151,13 @@ export namespace entity {
        * @type {string}
        */
       if (typeCastOptions) {
-        if (
-          typeCastOptions.integerTypeCastFunction &&
-          typeof typeCastOptions.integerTypeCastFunction !== 'function'
-        ) {
+        this.typeCastFunction = typeCastOptions.integerTypeCastFunction;
+        if (typeof typeCastOptions.integerTypeCastFunction !== 'function') {
           throw new Error(
-            `integerTypeCastFunction is not a function or was not provided.`
+            `integerTypeCastFunction is not a function or is not provided.`
           );
         }
-        this.typeCastFunction = typeCastOptions.integerTypeCastFunction;
+
         this.typeCastProperties = typeCastOptions.properties
           ? arrify(typeCastOptions.properties)
           : undefined;
@@ -167,19 +165,19 @@ export namespace entity {
     }
     // tslint:disable-next-line no-any
     valueOf(): any {
-      let customCast = this.typeCastFunction ? true : false;
+      let shouldCustomCast = this.typeCastFunction ? true : false;
       if (
         this.typeCastProperties &&
         !this.typeCastProperties.includes(this._entityPropertyName!)
       ) {
-        customCast = false;
+        shouldCustomCast = false;
       }
 
-      if (customCast) {
+      if (shouldCustomCast) {
         try {
           return this.typeCastFunction!(this.value);
         } catch (error) {
-          error.message = `integerTypeCastFunction threw an error - ${error.message}`;
+          error.message = `integerTypeCastFunction threw an error:\n\n - ${error.message}`;
           throw error;
         }
       } else {
@@ -455,7 +453,7 @@ export namespace entity {
           "Please consider passing 'options.wrapNumbers=true' and\n" +
           "'options.integerTypeCastOptions' as\n" +
           '{\n' +
-          '  integerTypeCastFunction: optionally provide <your_custom_function>\n' +
+          '  integerTypeCastFunction: provide <your_custom_function>\n' +
           '  properties: optionally specify property name(s) to be cutom casted' +
           '}\n' +
           'to prevent this error.'
@@ -469,10 +467,14 @@ export namespace entity {
    *
    * @private
    * @param {object} valueProto The protobuf Value message to convert.
-   * @param {object} typeCastOptions Config for custom `integerValue` cast.
-   * @param {boolean} [wrapNumbers=false] Indicates if the numbers should be
-   *     wrapped in Int wrapper.
-   * @param {function} [typeCastOptions.integerTypeCastFunction] A custom user
+   * @param {boolean} [wrapNumbers=false] Wrap values of integerValue type in 
+   *     {@link Datastore#Int} object.
+   * @param {object} [typeCastOptions] Configuration to convert
+   *     values of `integerValue` type to a custom value. Must provide an
+   *     `integerTypeCastFunction` to handle `integerValue` conversion.
+   *     Note: integerTypeCastOptions is ignored when `options.wrapNumbers` is
+   *     not set to `true`.
+   * @param {function} typeCastOptions.integerTypeCastFunction A custom user
    *     provided function to convert `integerValue`.
    * @param {sting|string[]} [typeCastOptions.properties] `Entity` property
    *     names to be converted using `integerTypeCastFunction`.
@@ -662,10 +664,14 @@ export namespace entity {
    *
    * @private
    * @param {object} entityProto The protocol entity object to convert.
-   * @param {object} typeCastOptions Config for custom `integerValue` cast.
-   * @param {boolean} [wrapNumbers=false] Indicates if the numbers should be
-   *     wrapped in Int wrapper.
-   * @param {function} [typeCastOptions.integerTypeCastFunction] A custom user
+   * @param {boolean} [wrapNumbers=false] Wrap values of integerValue type in 
+   *     {@link Datastore#Int} object.
+   * @param {object} [typeCastOptions] Configuration to convert
+   *     values of `integerValue` type to a custom value. Must provide an
+   *     `integerTypeCastFunction` to handle `integerValue` conversion.
+   *     Note: integerTypeCastOptions is ignored when `options.wrapNumbers` is
+   *     not set to `true`.
+   * @param {function} typeCastOptions.integerTypeCastFunction A custom user
    *     provided function to convert `integerValue`.
    * @param {sting|string[]} [typeCastOptions.properties] `Entity` property
    *     names to be converted using `integerTypeCastFunction`.
@@ -891,10 +897,14 @@ export namespace entity {
    * @param {object[]} results The response array.
    * @param {object} results.entity An entity object.
    * @param {object} results.entity.key The entity's key.
-   * @param {object} typeCastOptions Config for custom `integerValue` cast.
-   * @param {boolean} [wrapNumbers=false] Indicates if the numbers should be
-   *     wrapped in Int wrapper.
-   * @param {function} [typeCastOptions.integerTypeCastFunction] A custom user
+   * @param {boolean} [wrapNumbers=false] Wrap values of integerValue type in 
+   *     {@link Datastore#Int} object.
+   * @param {object} [typeCastOptions] Configuration to convert
+   *     values of `integerValue` type to a custom value. Must provide an
+   *     `integerTypeCastFunction` to handle `integerValue` conversion.
+   *     Note: integerTypeCastOptions is ignored when `options.wrapNumbers` is
+   *     not set to `true`.
+   * @param {function} typeCastOptions.integerTypeCastFunction A custom user
    *     provided function to convert `integerValue`.
    * @param {sting|string[]} [typeCastOptions.properties] `Entity` property
    *     names to be converted using `integerTypeCastFunction`.
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -63,8 +63,14 @@ export namespace entity {
    * const aDouble = datastore.double(7.3);
    */
   export class Double {
+    type: string;
     value: number;
     constructor(value: number) {
+      /**
+       * @name Double#type
+       * @type {string}
+       */
+      this.type = 'DatastoreDouble';
       /**
        * @name Double#value
        * @type {number}
@@ -85,6 +91,23 @@ export namespace entity {
   }
 
   /**
+   * Check if a value is a Datastore Double object converted from JSON.
+   *
+   * @private
+   * @param {*} value
+   * @returns {boolean}
+   */
+  export function isDsDoubleLike(value: unknown) {
+    const maybeDsDouble = value as Double;
+    return (
+      isDsDouble(maybeDsDouble) ||
+      (is.object(maybeDsDouble) &&
+        is.number(maybeDsDouble.value) &&
+        maybeDsDouble.type === 'DatastoreDouble')
+    );
+  }
+
+  /**
    * Build a Datastore Int object. For long integers, a string can be provided.
    *
    * @class
@@ -101,6 +124,7 @@ export namespace entity {
    * const anInt = datastore.int(7);
    */
   export class Int extends Number {
+    type: string;
     value: string;
     typeCastFunction?: Function;
     typeCastProperties?: string[];
@@ -116,6 +140,11 @@ export namespace entity {
         typeof value === 'object'
           ? value.integerValue.toString()
           : value.toString();
+      /**
+       * @name Int#type
+       * @type {string}
+       */
+      this.type = 'DatastoreInt';
       /**
        * @name Int#value
        * @type {string}
@@ -168,6 +197,23 @@ export namespace entity {
    */
   export function isDsInt(value?: {}) {
     return value instanceof entity.Int;
+  }
+
+  /**
+   * Check if a value is a Datastore Int object converted from JSON.
+   *
+   * @private
+   * @param {*} value
+   * @returns {boolean}
+   */
+  export function isDsIntLike(value: unknown) {
+    const maybeDsInt = value as Int;
+    return (
+      isDsInt(maybeDsInt) ||
+      (is.object(maybeDsInt) &&
+        is.string(maybeDsInt.value) &&
+        maybeDsInt.type === 'DatastoreInt')
+    );
   }
 
   export interface Coordinates {
@@ -308,7 +354,11 @@ export namespace entity {
       if (options.path.length % 2 === 0) {
         const identifier = options.path.pop();
 
-        if (is.number(identifier) || isDsInt(identifier)) {
+        if (
+          is.number(identifier) ||
+          isDsInt(identifier) ||
+          isDsIntLike(identifier)
+        ) {
           this.id = (((identifier as {}) as Int).value || identifier) as string;
         } else if (is.string(identifier)) {
           this.name = identifier as string;
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -113,6 +113,8 @@ export namespace entity {
    * @class
    * @param {number|string} value The integer value.
    * @param {object} typeCastOptions Config for custom `integerValue` cast.
+   * @param {boolean} [wrapNumbers=false] Indicates if the numbers should be
+   *     wrapped in Int wrapper.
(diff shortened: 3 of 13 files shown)


AVaksman:
@bcoe please take a look.
Looking at both approaches in this PR and #488, I think this PR came a bit over complex than this feature needs to be 😄. Now I am leaning towards the [first option](https://github.com/googleapis/nodejs-datastore/pull/488#discussion_r326279084) and see add more in the future if there is a need :relaxed:

src/entity.ts
```diff
@@ -97,12 +102,68 @@ export namespace entity {
    */
   export class Int {
```
stephenplusplus: Can we extend `Number` here?

stephenplusplus:
Should we think about forcing a user to opt-in or opt-out of the new throwing behavior? In Spanner, by default, the user will receive the Int objects. They can call [`row.toJSON()`](https://github.com/googleapis/nodejs-spanner/blob/76d17b09720de5e70eeaa91bfc2ede1875e4c122/system-test/spanner.ts#L302-L330) on the response from a query, which tries to turn everything into native values (e.g., `Number(8)` instead of `new Int(8)`). When a number is out of bounds, it will throw at this time.

The user also has the choice to provide an option, [`row.toJSON({wrapNumbers:true})`](https://github.com/googleapis/nodejs-spanner/blob/76d17b09720de5e70eeaa91bfc2ede1875e4c122/system-test/spanner.ts#L332), which will return the custom Int type, deferring any throwing until the user tries to use the out of bounds number.

In Datastore terms, we could:

  1. Always try to return the normal `Number(int)` value (as opposed to `new DSInt(int)`), and error/throw during the `.get()` only when a number is out of bounds
  2. Allow an option to `get` (and anywhere else) that allows `wrapNumbers: true`. We would always return the custom Int class, which would defer throwing until the user tries to use an out of bounds value

I like this approach, as it would make this breaking change more practical for users in both camps-- where out of bounds integers aren't relevant (no errors), as well as where they are (useful errors). It's almost a fix instead of a breaking change at that point, but I won't argue for that :)

src/entity.ts
```diff
@@ -89,6 +89,11 @@ export namespace entity {
    *
    * @class
    * @param {number|string} value The integer value.
+   * @param {object} integerTypeCastOptions Config for custom `integerValue` cast.
+   * @param {function} [integerTypeCastOptions.integerTypeCastFunction] A custom user
+   *     provided function to convert `integerValue`.
+   * @param {sting|string[]} [integerTypeCastOptions.names] `Entity` property
```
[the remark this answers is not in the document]
AVaksman: amended to `properties`

src/entity.ts
```diff
@@ -97,12 +102,68 @@ export namespace entity {
    */
   export class Int {
     value: string;
-    constructor(value: number | string) {
+    integerTypeCastOptions?: IntegerTypeCastOptions;
+    // private _numericValue: number | string;
```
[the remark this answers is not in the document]
AVaksman: removed, thanks

src/entity.ts
```diff
@@ -97,12 +102,68 @@ export namespace entity {
...
+      this.value =
+        typeof value === 'object'
+          ? value.integerValue.toString()
+          : value.toString();
       /**
        * @name Int#value
        * @type {string}
        */
-      this.value = value.toString();
+      this.integerTypeCastOptions = integerTypeCastOptions;
```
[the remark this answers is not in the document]
AVaksman: Refactored accordingly

src/entity.ts
```diff
@@ -97,12 +102,68 @@ export namespace entity {
...
+          typeof this.integerTypeCastOptions!.integerTypeCastFunction !==
+          'function'
+        ) {
+          throw new Error(
+            `integerTypeCastFunction is not a function or was not provided.`
+          );
+        }
+        customCast = true;
+        if (this.integerTypeCastOptions.names) {
+          this.integerTypeCastOptions.names = arrify(
```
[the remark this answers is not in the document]
AVaksman: Yep

src/entity.ts
```diff
@@ -505,7 +577,11 @@ export namespace entity {
    *
    * @private
    * @param {object} entityProto The protocol entity object to convert.
-   * @returns {object}
+   * @param {object} integerTypeCastOptions Config for custom `integerValue` cast.
+   * @param {function} [integerTypeCastOptions.integerTypeCastFunction] A custom user
+   *     provided function to convert `integerValue`.
+   * @param {sting|string[]} [integerTypeCastOptions.names] `Entity` property
+   *     names to be converted using `integerTypeCastFunction`.* @returns {object}
```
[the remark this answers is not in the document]
AVaksman: Oops, fixed

src/entity.ts
AVaksman: Done

test/entity.ts
```diff
@@ -66,6 +67,102 @@ describe('entity', () => {
...
+      beforeEach(() => {
+        valueProto = {
+          valueType: 'integerValue',
+          integerValue: 8,
+        };
+      });
+
+      it('should throw if integerTypeCastFunction is not provided', () => {
+        assert.throws(
+          () => new entity.Int(valueProto, {}).valueOf(),
```
bcoe: should we default to providing a casting function that throws if the integer is larger than max int?

bcoe: (seems like `valueOf` should work by default).

samples/concepts.js
```diff
@@ -1018,11 +1018,11 @@ class Transaction extends TestHelper {
       // Restore `datastore` to the mock API.
       datastore = datastoreMock;
       assert.strictEqual(
-        accounts[0].balance,
+        accounts[0].balance.valueOf(),
```
bcoe: Could we add a couple examples in:

https://github.com/googleapis/nodejs-datastore/blob/master/samples/concepts.js#L272

That demonstrate:

1. how to save an entry with an integer value (I believe we already have this).
2. how to fetch an entry with an integer value.
3. potentially how we'd JSON stringify an entry we fetched?

I know we're already showing the conversion with `accounts[0].balance.valueOf()`, but I think a more self contained example might be nice.

samples/concepts.js
```diff
@@ -1018,11 +1018,11 @@ class Transaction extends TestHelper {
       // Restore `datastore` to the mock API.
       datastore = datastoreMock;
       assert.strictEqual(
```
bcoe: could we also add a `assert.equal` test that demonstrates `valueOf` gets called automatically? will it not get called in a `strictEqual` check?

test/entity.ts
crwilcox: +1. This seems good. It makes the default the same behavior of the original except we throw on corruption of the data.

crwilcox: Looking closer it seems this is just validating if an empty dict is given as a param it fails as that isn't valid.

samples/concepts.js
crwilcox: added additional checks. resolving.

samples/concepts.js
crwilcox: I am not sure this can be done without looking odd. This sample exists alongside other language ones that are this simple.

I would be fine making a future work issue for this. 

/cc: @BenWhitehead

BenWhitehead: Since we don't currently have any section in the docs discussing how to work with "large" numbers, and I'm not sure when we will. I think it'd be good to at least have a commented sample in the repo that can be discovered and referred to independent from this development PR. Once we have the sample we can add a work item to get it published into the docs.

system-test/datastore.ts
```diff
@@ -57,11 +57,11 @@ describe('Datastore', () => {
       publishedAt: new Date(),
       author: 'Silvano',
       isDraft: false,
-      wordCount: 400,
-      rating: 5.0,
+      wordCount: datastore.int(400),
+      rating: datastore.int(5.0),
```
bcoe: this feels a bit weird, because `5.0` looks like a floating point. How do we handle floating point values in this new world?

samples/concepts.js
crwilcox: I am capturing this in an internal bug b/14309461.

If you feel strongly this is needed, it may belong in a doc. comment with the function being called?

BenWhitehead: As long as there is something (either in samples, or in the reference docs alongside the code) so that someone can learn about this, my main concern is satisfied. Since this is a bit of an odd compatibility case that not all of the supported languages necessitate I think it'll take some design/work to incorporate into the docs based on how they're structured right now.

src/entity.ts
```diff
@@ -89,20 +89,73 @@ export namespace entity {
...
+          );
+        }
+        this.typeCastFunction = typeCastOptions.integerTypeCastFunction;
+        this.typeCastProperties = typeCastOptions.properties
+          ? arrify(typeCastOptions.properties)
+          : undefined;
+      }
+    }
+    // tslint:disable-next-line no-any
+    valueOf(): any {
```
jkwlui: We can make this type-safe for typescript users by making this class generic on the return type of `valueOf()`:

```
class Int<T> extends Number ...
```

We'll also be able to make `integerTypeCastFunction` type-safe by making `IntegerTypeCastOptions` generic:

```typescript
export interface IntegerTypeCastOptions<T> {
  integerTypeCastFunction: (value: string) => T;
  properties?: string | string[];
}
```

jkwlui: Nevermind folks.. It's not possible to override the return type of `valueOf()` to the generic type since `Int` inherits from `Number`. Typescript really dislike type-hacking..

jkwlui: This also means Typescript users would have to coerce type returned by `valueOf()`:

```typescript
// assume integerTypeCastFunction returns a BigInt
const bigInt = myInt.valueOf() as BigInt;
```

test/entity.ts
```diff
@@ -167,15 +264,17 @@ describe('entity', () => {
         namespace: 'namespace',
         path: ['ParentKind', 'name', 'Kind', 1, 'SubKind', new entity.Int('1')],
       });
+      const num = new entity.Int(1).valueOf();
```
jkwlui: `num` and `str` are not used (perhaps you meant to replace the items in the `path` array below?

AVaksman:
@bcoe @crwilcox @stephenplusplus PTAL
Refactored to add `wrapNumbers` param  as a part of `IntegerTypeCastOptions`
Behavior after last changes.

- On default 
   - library will try to convert to native `Number`
   - Will throw on overflow.
- User may pass `{wrapNumbers: true}` as part of `options.integerTypeCastOptions`
   - Library will return DsInt object.
     1. If `integerTypeCastFunction` is not provided
         - on `@valueOf` the library will try to convert to native `Number`
         - Will throw on overflow.
         - `#value` will return string representation of the integer value
     2. If `integerTypeCastFunction` is provided `
         - on `@valueOf` the library will return the result of the `integerTypeCastFunction`
         - `#value` will return string representation of the integer value

src/request.ts
```diff
@@ -763,7 +783,10 @@ class DatastoreRequest {
       let entities: Entity[] = [];
 
       if (resp.batch.entityResults) {
-        entities = entity.formatArray(resp.batch.entityResults);
+        entities = entity.formatArray(
```
stephenplusplus: Maybe here would be a good place to really friendlify the error message, so that if any error was thrown, it could say “One of the values was out of bounds. To avoid this...” etc.

AVaksman: did you mean friendlify this message?
https://github.com/googleapis/nodejs-datastore/pull/516/files/a2eb3e3f1f3ff8c222cc1e8e078932db56886c5e#diff-f6515d62cc6523369af544f6e8852d1cR447

stephenplusplus: Yes! If we can put an error in the right place (like when calling get()), the user would get a message that explains that we tried to return all of the results, but the value for "phoneNumber" was out of bounds. It could also explain to use `wrapNumbers` to circumvent the error.

src/entity.ts
```diff
@@ -376,11 +437,44 @@ export namespace entity {
...
+   */
+  function decodeIntegerValue(value: ValueProto) {
+    const num = Number(value.integerValue);
+    if (!Number.isSafeInteger(num)) {
+      throw new Error(
+        'We attempted to return all of the numeric values, but ' +
+          (value.propertyName ? value.propertyName + ' ' : '') +
+          'value ' +
+          value.integerValue +
+          ' is out of bounds of Number safe integer.\n' +
```
stephenplusplus: Could you replace "Number safe integer" with "Number.MAX_SAFE_INTEGER"? It will give people something to Google if they're unfamiliar.

src/request.ts
```diff
@@ -580,6 +591,15 @@ class DatastoreRequest {
    *     [here](https://cloud.google.com/datastore/docs/articles/balancing-strong-and-eventual-consistency-with-google-cloud-datastore).
    * @param {object} [options.gaxOptions] Request configuration options, outlined
    *     here: https://googleapis.github.io/gax-nodejs/global.html#CallOptions.
+   * @param {object} [options.integerTypeCastOptions] Configurations to
+   *     optionally wrap `integerValue` in Datastore Int object and optionally
+   *     provide an `integerTypeCastFunction` to handle `integerValue` conversion
+   * @param {boolean} [options.integerTypeCastOptions.wrapnumbers=false]
```
stephenplusplus:
```diff
- wrapnumbers
+ wrapNumbers
```

stephenplusplus: Could we promote or alias `wrapNumbers` one level up on the `options` object (`options.wrapNumbers`)? Using this should opt the user out of the custom functionality of `integerTypeCastFunction`, as it's the simplest option to have number safety.

src/entity.ts
AVaksman: Done

src/request.ts
AVaksman: Thanks 👍

crwilcox: I think I would like to wait to see if that is necessary before cluttering things up?

@bcoe

src/request.ts
crwilcox: Is this done? This message seems pretty friendly. https://github.com/googleapis/nodejs-datastore/pull/516/files/c4d944e6503221f685bb6499e34554e0107d0d3f#diff-b6ce7cf07299e63c027f59d103d82e81R109

src/request.ts
stephenplusplus: We talked about this in the chat, @AVaksman was going to make it work. I don't see the clutter though-- focusing on ease of use for the user, "options.wrapNumbers" is easier to remember and more intuitive than "options.integerTypeCastOptions = { wrapNumbers: true }".

src/entity.ts
```diff
@@ -111,16 +111,36 @@ export namespace entity {
    *
    * @class
    * @param {number|string} value The integer value.
+   * @param {object} typeCastOptions Config for custom `integerValue` cast.
+   * @param {boolean} [wrapNumbers=false] Indicates if the numbers should be
```
stephenplusplus: I think this line can be removed here.

src/entity.ts
```diff
@@ -130,7 +150,48 @@ export namespace entity {
...
        * @type {string}
        */
-      this.value = value.toString();
+      if (typeCastOptions) {
+        if (
+          typeCastOptions.integerTypeCastFunction &&
+          typeof typeCastOptions.integerTypeCastFunction !== 'function'
+        ) {
+          throw new Error(
+            `integerTypeCastFunction is not a function or was not provided.`
```
stephenplusplus: This is a small thing, but it doesn't look possible for this error to be thrown when `integerTypeCastFunction` was not provided. Maybe just this would work:

```js
this.typeCastFunction = typeCastOptions.integerTypeCastFunction;

if (typeof this.typeCastFunction !== 'function') {
  throw ...
}
```

src/entity.ts
```diff
@@ -130,7 +150,48 @@ export namespace entity {
...
+          );
+        }
+        this.typeCastFunction = typeCastOptions.integerTypeCastFunction;
+        this.typeCastProperties = typeCastOptions.properties
```
stephenplusplus: I think it's fair to just do `this.typeCastProperties = arrify(typeCastOptions.properties)`, even if it's empty 🤷‍♂

src/entity.ts
```diff
@@ -130,7 +150,48 @@ export namespace entity {
...
+          ? arrify(typeCastOptions.properties)
+          : undefined;
+      }
+    }
+    // tslint:disable-next-line no-any
+    valueOf(): any {
+      let customCast = this.typeCastFunction ? true : false;
```
stephenplusplus: It might be a bit more readable to ask exactly what we're wondering:

```js
let shouldCustomCast = typeof this.typeCastFunction === 'function';
```

src/entity.ts
```diff
@@ -130,7 +150,48 @@ export namespace entity {
...
+        !this.typeCastProperties.includes(this._entityPropertyName!)
+      ) {
+        customCast = false;
+      }
+
+      if (customCast) {
+        try {
+          return this.typeCastFunction!(this.value);
+        } catch (error) {
+          error.message = `integerTypeCastFunction threw an error - ${error.message}`;
```
stephenplusplus: Up for debate, but newlining here might print better:

```js
error.message = `integerTypeCastFunction returned an error:\n\n  ${error.message}`;
```

src/entity.ts
```diff
@@ -376,11 +437,45 @@ export namespace entity {
...
+  function decodeIntegerValue(value: ValueProto) {
+    const num = Number(value.integerValue);
+    if (!Number.isSafeInteger(num)) {
+      throw new Error(
+        'We attempted to return all of the numeric values, but ' +
+          (value.propertyName ? value.propertyName + ' ' : '') +
+          'value ' +
+          value.integerValue +
+          " is out of bounds of 'Number.MAX_SAFE_INTEGER'.\n" +
+          "Please consider passing 'options.wrapNumbers=true' and\n" +
```
stephenplusplus:
```diff
- and\n" +
+ or\n" +
```

?

src/entity.ts
```diff
@@ -376,11 +437,45 @@ export namespace entity {
...
+          "'options.integerTypeCastOptions' as\n" +
+          '{\n' +
+          '  integerTypeCastFunction: optionally provide <your_custom_function>\n' +
```
stephenplusplus: I would lose `optionally` here from the description, because once you've opted in to `integerTypeCastOptions`, this is no longer optional.

src/entity.ts
```diff
@@ -376,11 +437,45 @@ export namespace entity {
...
+    }
+    return num;
+  }
+
   /**
    * Convert a protobuf Value message to its native value.
    *
    * @private
    * @param {object} valueProto The protobuf Value message to convert.
+   * @param {object} typeCastOptions Config for custom `integerValue` cast.
```
stephenplusplus: I think this line needs to drop one down.

stephenplusplus: I think `typeCastOptions` is optional, so it should be bracketed. And then `typeCastOptions.integerTypeCastFunction` is required within that object, so I don't think it should be bracketed. I'm not sure that's true, though!

src/entity.ts
```diff
@@ -554,6 +662,13 @@ export namespace entity {
    *
    * @private
    * @param {object} entityProto The protocol entity object to convert.
+   * @param {object} typeCastOptions Config for custom `integerValue` cast.
```
stephenplusplus: Drop one line, and same concerns about where to bracket to denote optionality.

src/entity.ts
```diff
@@ -768,7 +891,13 @@ export namespace entity {
    * @param {object[]} results The response array.
    * @param {object} results.entity An entity object.
    * @param {object} results.entity.key The entity's key.
-   * @returns {object[]}
+   * @param {object} typeCastOptions Config for custom `integerValue` cast.
```
stephenplusplus: This is a comment about line dropping and brackets.

src/query.ts
```diff
@@ -397,6 +397,19 @@ class Query {
    *     If not specified, default values are chosen by Datastore for the
    *     operation. Learn more about strong and eventual consistency
    *     [here](https://cloud.google.com/datastore/docs/articles/balancing-strong-and-eventual-consistency-with-google-cloud-datastore).
+   * @param {object} [options.gaxOptions] Request configuration options, outlined
+   *     here: https://googleapis.github.io/gax-nodejs/global.html#CallOptions.
+   * @param {boolean} [options.wrapNumbers=false]
+   *     Indicates if the numbers should be wrapped in Int wrapper.
+   * @param {object} [options.integerTypeCastOptions] Configurations to
+   *     optionally wrap `integerValue` in Datastore Int object and optionally
```
stephenplusplus: I think the first sentence can be changed to.

Configuration to convert values of `integerValue` type to a custom value.

src/query.ts
```diff
@@ -397,6 +397,19 @@ class Query {
...
+   *     provide an `integerTypeCastFunction` to handle `integerValue` conversion.
+   *     Note: `integerTypeCastingOptions` values will be ingnored
```
stephenplusplus: Note: `integerTypeCastOptions` is ignored when `options.wrapNumbers` is set.

src/query.ts
```diff
@@ -397,6 +397,19 @@ class Query {
    *     If not specified, default values are chosen by Datastore for the
    *     operation. Learn more about strong and eventual consistency
    *     [here](https://cloud.google.com/datastore/docs/articles/balancing-strong-and-eventual-consistency-with-google-cloud-datastore).
+   * @param {object} [options.gaxOptions] Request configuration options, outlined
+   *     here: https://googleapis.github.io/gax-nodejs/global.html#CallOptions.
+   * @param {boolean} [options.wrapNumbers=false]
+   *     Indicates if the numbers should be wrapped in Int wrapper.
```
stephenplusplus: `Wrap values of `integerValue` type in {@Datastore#Int} objects.`

^-- I'm not actually sure if {@Datastore#Int} is the right way to link to that definition?

src/request.ts
```diff
@@ -432,6 +436,17 @@ class DatastoreRequest {
    *     [here](https://cloud.google.com/datastore/docs/articles/balancing-strong-and-eventual-consistency-with-google-cloud-datastore).
    * @param {object} [options.gaxOptions] Request configuration options, outlined
    *     here: https://googleapis.github.io/gax-nodejs/global.html#CallOptions.
+   * @param {boolean} [options.wrapNumbers=false]
```
stephenplusplus: Same notes as above throughout this documentation block.

src/request.ts
```diff
@@ -580,6 +594,17 @@ class DatastoreRequest {
...
```
stephenplusplus: Same notes about documentation for this and the properties that follow.

src/entity.ts
AVaksman: Done

src/entity.ts
AVaksman: Done

src/query.ts
AVaksman: Done

src/query.ts
AVaksman: It seems it should be {@link Datastore#Int} at least that's what the JSDoc documentation says

src/request.ts
AVaksman: Done

src/request.ts
AVaksman: Done

src/query.ts
AVaksman: @crwilcox I added a few tests in the commit [7c366cc](https://github.com/googleapis/nodejs-datastore/pull/516/commits/7c366ccf20f7421eaa8a750302e8165e8354eb8d) to test the "ignore `integerTypeCastOptions` if `options.wrapNumbers` is not set to `true`"

src/entity.ts
stephenplusplus: Oh, I see. I thought integerTypeCast options was its own thing. I’m not sure that I think we should require both wrapNumbers to be true as well as the full iTC object, since providing an iTC object should be enough to imply they want to handle conversion themselves.

Terminology wise, wrapping numbers is something that Datastore.Int does, but it’s not necessarily what the user function will do. I think they should be separate options the user has to process numbers, as opposed to being linked.

AVaksman: The current theme is as follows:
1. Default not passing anything (not wrapping in DsInt) - API tries to convert to `Number`, throws on overflow.
    In case of an exception, user has to make another network call to retrieve the value.
2. Pass just `wrapNumbers=true` - API will return a `DsInt` obj in which
    - `#valueOf` will act the same way as default behavior, i.e. try to convert to `Number`, throw on overflow. However now, in case of an exception with `#valueOf` user can still retrieve a string representation of the value with `#value` and does not need to make another network call. That is described as "minimal number safety"
3. Along with `wrapNumbers=true` user also passing `iTCFunction` - API will return `DsInt` obj in which.
   - `#valueOf` API will use `iTCFunction` to convert the numeric values.

User's function is only referred to by `#valueOf` method which is DsInt component. That is why numbers need to be wrapped to be handled by `iTCFunction`. 

4. If user just sets `iTCFunction`, but not `wrapNumbers=true` - API will treat it as default behavior.

src/entity.ts
AVaksman: Done

crwilcox:
@stephenplusplus ptal.

src/entity.ts
stephenplusplus: We had a follow-up conversation about this in the chat— I had some suggestions, but I’m not sure you replied after that!

src/entity.ts
stephenplusplus: The first reason I thought to suggest this change was due to the `this.typeCastFunction ? true : false`, which is somewhat unusual-- that is coercing the type based on if it's not undefined/null. For readability, I would rather see it say exactly what we're wondering: "We should custom cast the integer if the user gave us a function", vs "We should custom cast the integer if the user gave us a non-null argument".

AVaksman:
@stephenplusplus 
Latest modifications

combine `wrapNumbers: bool` and `iTCO: iTCO` parameters into one `wrapNumbersOptions: boolean | iTCO` (defaults to `false`) param.

With new changes user can
1. Default - `query.run(q)`
    - API will try to convert to `Number`
    - Will throw on overflow
2. Pass `options.wNO = true` (just wrap numbers, no custom cast) - `query.run(q, {wrapNumbersOptions: true})`
    - API will return a DsInt obj in which
      - #valueOf will act the same way as default behavior, i.e. try to convert to Number, throw on overflow. However now, in case of an exception with #valueOf user can still retrieve a string representation of the numerical value with `#value` and does not need to make another network call.
3. Pass `options.wNO = iTCO` - `query.run(q, {wrapNumbersOptions: {iTCF = myFunction}})`
    - API will return a DsInt obj in which
      - `#valueOf` API will use iTCF to convert the numeric values.
      - `#value` still returns the string representation of the numerical value

PTAL

src/entity.ts
```diff
@@ -376,11 +435,49 @@ export namespace entity {
...
+  function decodeIntegerValue(value: ValueProto) {
+    const num = Number(value.integerValue);
+    if (!Number.isSafeInteger(num)) {
+      throw new Error(
+        'We attempted to return all of the numeric values, but ' +
+          (value.propertyName ? value.propertyName + ' ' : '') +
+          'value ' +
+          value.integerValue +
+          " is out of bounds of 'Number.MAX_SAFE_INTEGER'.\n" +
+          "Please consider passing 'options.wrapNumbersOptions=true' or\n" +
```
stephenplusplus: I like that you went with a dual-purpose option, where it can be a boolean, for simple mode, or an object for an advanced use case. I would rather have `wrapNumbers` be the name, however, so that `wrapNumbers: true` is possible, compared to `wrapNumbersOptions: true`. When the object is used, `wrapNumbers = {integerTypeCastFunction: () => {}}` still seems fine.

src/entity.ts
```diff
@@ -376,11 +435,49 @@ export namespace entity {
...
+      );
+    }
+    return num;
+  }
+
+  /**
+   * @typedef {object} IntegerTypeCastOptions Configuration to convert
+   *     values of `integerValue` type to a custom value. Must provide an
+   *     `integerTypeCastFunction` to handle `integerValue` conversion.
+   * @property {function} integerTypeCastFunction A custom user
```
stephenplusplus: I think users will be surprised when they learn `integerValue`s are not converted to what they returned from their function until "valueOf()" is called. Imagining a scenario where a user has their own Int implementation, `AppBigInt`, I think they would expect `instanceof entity.data.myNumber === AppBigInt` rather than our `DatastoreInt`-- which they probably assumed they were opting out of.

If you still think having it work the way it does in this PR now would be best, let's beef up the docs to over-explain how their return value will come into play.

src/query.ts
```diff
@@ -397,6 +397,10 @@ class Query {
    *     If not specified, default values are chosen by Datastore for the
    *     operation. Learn more about strong and eventual consistency
    *     [here](https://cloud.google.com/datastore/docs/articles/balancing-strong-and-eventual-consistency-with-google-cloud-datastore).
+   * @param {object} [options.gaxOptions] Request configuration options, outlined
+   *     here: https://googleapis.github.io/gax-nodejs/global.html#CallOptions.
+   * @param {boolean | IntegerTypeCastOptions} [options.wrapNumbersOptions=false]
```
stephenplusplus: The description should explain "If a boolean, this will wrap values... If an object, you can customize the behavior..."

src/entity.ts
stephenplusplus: Using the logic flow you wrote above, this is my proposal:

1. Default - `query.run(q)`
    - API will try to convert to `Number`
    - Will throw on overflow
2. Pass options.wrapNumbers as boolean (just wrap numbers, no custom cast) - `query.run(q, {wrapNumbers: true})`
    - API will return a DsInt obj in which
      - #valueOf will act the same way as default behavior, i.e. try to convert to Number, throw on overflow. However now, in case of an exception with #valueOf user can still retrieve a string representation of the numerical value with `#value` and does not need to make another network call.
3. Pass options.wrapNumbers as object - `query.run(q, {wrapNumbers: {integerTypeCastFunction: = myFunction}})`
    - API will return the return value from user-provided `wrapNumbers.integerTypeCastFunction`

src/query.ts
AVaksman: Does below sounds ok?
```
   * @param {boolean | IntegerTypeCastOptions} [options.wrapNumbers=false]
   *     Wrap values of integerValue type in {@link Datastore#Int} object.
   *     If a `boolean`, this will wrap values in {@link Datastore#Int}.
   *     If an `object`, this will return a value returned by 
   *     `wrapNumbers.integerTypeCastFunction`.
   *     Please see {@link  IntegerTypeCastOptions} for options descriptions.

```

src/entity.ts
AVaksman: Renamed

src/entity.ts
AVaksman: Implemented

src/query.ts
AVaksman: Pushed the changes 👍

package.json
```diff
@@ -58,7 +58,8 @@
     "google-gax": "^1.7.5",
     "is": "^3.3.0",
     "split-array-stream": "^2.0.0",
-    "stream-events": "^1.0.5"
+    "stream-events": "^1.0.5",
+    "typescript": "~3.7.0"
```
stephenplusplus: Do we need this?

src/entity.ts
```diff
@@ -376,11 +435,53 @@ export namespace entity {
...
+  function decodeIntegerValue(value: ValueProto) {
+    const num = Number(value.integerValue);
+    if (!Number.isSafeInteger(num)) {
+      throw new Error(
+        'We attempted to return all of the numeric values, but ' +
+          (value.propertyName ? value.propertyName + ' ' : '') +
+          'value ' +
+          value.integerValue +
+          " is out of bounds of 'Number.MAX_SAFE_INTEGER'.\n" +
+          "Please consider passing 'options.wrapNumbers=true' or\n" +
```
stephenplusplus: Could you put the “To prevent this error, please...”. That way, it doesn’t float on its own line after the demonstration object.

src/entity.ts
```diff
@@ -130,7 +150,46 @@ export namespace entity {
...
+        !this.typeCastProperties.includes(this._entityPropertyName!)
+      ) {
+        shouldCustomCast = false;
+      }
+
+      if (shouldCustomCast) {
+        try {
+          return this.typeCastFunction!(this.value);
+        } catch (error) {
+          error.message = `integerTypeCastFunction threw an error:\n\n - ${error.message}`;
```
stephenplusplus: Could you do two spaces in front of the hyphen? I think that’s the standard newline+indent formatting we use in other places.

src/entity.ts
```diff
@@ -376,11 +435,53 @@ export namespace entity {
...
+   * @property {string | string[]} [properties] `Entity` property
+   *     names to be converted using `integerTypeCastFunction`.
+   */
   /**
    * Convert a protobuf Value message to its native value.
    *
    * @private
    * @param {object} valueProto The protobuf Value message to convert.
+   * @param {boolean | IntegerTypeCastOptions} [wrapNumbers=false] Wrap values of integerValue type in
+   *     {@link Datastore#Int} object.
```
stephenplusplus: “objects.”

src/entity.ts
```diff
@@ -376,11 +435,53 @@ export namespace entity {
...
+   *     If a `boolean`, this will wrap values in {@link Datastore#Int}.
```
stephenplusplus: “objects.” at the end there please!

src/entity.ts
```diff
@@ -376,11 +435,53 @@ export namespace entity {
...
+   *     If an `object`, this will return a value returned by
+   *     `wrapNumbers.integerTypeCastFunction`.
+   *     Please see {@link  IntegerTypeCastOptions} for options descriptions.
```
stephenplusplus: I’m not sure this will make a difference, but there are two spaces before ‘IntegerTypeCastOptions’.

test/entity.ts
```diff
@@ -237,33 +385,269 @@ describe('entity', () => {
...
+          entity.decodeValueProto(valueProto, wrapNumbers);
+          assert.strictEqual(stub.called, true);
+        });
+
+        it('should call #valueOf if integerTypeCastFunction is provided', () => {
+          const stub = sinon.stub();
+          const wrapNumbers = {integerTypeCastFunction: stub};
+
+          entity.decodeValueProto(valueProto, wrapNumbers);
+          assert.strictEqual(stub.called, true);
```
stephenplusplus: Let’s also check that the value returned is what the iTCF returned.

src/entity.ts
```diff
@@ -130,7 +150,46 @@ export namespace entity {
        * @name Int#value
        * @type {string}
        */
-      this.value = value.toString();
+      if (typeCastOptions) {
+        this.typeCastFunction = typeCastOptions.integerTypeCastFunction;
+        if (typeof typeCastOptions.integerTypeCastFunction !== 'function') {
+          throw new Error(
+            `integerTypeCastFunction is not a function or is not provided.`
```
stephenplusplus: “was not provided.”

test/entity.ts
```diff
@@ -237,33 +385,269 @@ describe('entity', () => {
...
+        ) => {
+          assert.strictEqual(wrapNumbers, undefined);
+          assert.strictEqual(entityProto, expectedValue);
+          return expectedValue;
+        };
+
+        assert.strictEqual(entity.decodeValueProto(valueProto), expectedValue);
+      });
+
+      it('should wrap nubers with an option', () => {
```
stephenplusplus: “numbers”

test/entity.ts
```diff
@@ -237,33 +385,269 @@ describe('entity', () => {
...
+
+        entity.entityFromEntityProto = (entityProto: {}) => {
+          assert.strictEqual(entityProto, expectedValue);
+          return expectedValue;
+        };
+
+        assert.strictEqual(entity.decodeValueProto(valueProto), expectedValue);
+      });
+
+      it('should not wrap nubers by default', () => {
```
stephenplusplus: “numbers”

test/request.ts
```diff
@@ -456,6 +456,53 @@ describe('Request', () => {
           .emit('reading');
       });
 
+      it('should pass `wrapNumbersOprions` to formatArray', done => {
```
stephenplusplus: I think these tests should be rewritten to assure all the asserts are called. If necessary, break them apart into multiple tests for each value type.

package.json
AVaksman: Accident, removed

src/entity.ts
AVaksman: Done

src/entity.ts
AVaksman: Done

src/entity.ts
AVaksman: Fixed

src/entity.ts
AVaksman: Fixed!

src/entity.ts
AVaksman: Fixed!

test/entity.ts
AVaksman: added functionality to `integerTypeCastFunction`
added an assert to verify `entity.decodeValueProto` returnes the value returned by `integerTypeCastFunction`

src/entity.ts
AVaksman: FIxed!

test/entity.ts
AVaksman: Fixed!

test/entity.ts
AVaksman: Fixed as well!

test/request.ts
AVaksman: Split all these tests into single test per each value.

src/entity.ts
AVaksman: Got it.
My line of thought was, since we already weeded out anything but a `function`, at this point the value of `this.typeCastFunction` can only be a `function`.
But for code readability purposes it makes sense.
Fixed it.

AVaksman:
@stephenplusplus 
I believe I addressed all the most resent comments PTAL

stephenplusplus:
Thank you very much for your patience with my reviews! It looks great to me!

Issue #588, opened 2020-02-10
decodeIntegerValue throws an error, but it's not caught.


jamie-pate:
#### Environment details

  - OS: Linux
Distributor ID: Ubuntu
Description:    Pop!_OS 19.10
Release:        19.10
Codename:       eoan
  - Node.js version: v10.15.2, v12.14.0
  - npm version: 6.4.1, 6.13.4, yarn 1.19.1
  - `@google-cloud/datastore` version: ^5.0.2 (5.0.3, 5.0.4)

#### Steps to reproduce

Add an integer value into the datastore that won't fit within the bounds of Number.MAX_SAFE_INTEGER

Try to request the data back from the datastore:
```JavaScript
const datastore = new Datastore({...})
const q = datastore.createQuery('sesionState').filter('event', '=', event || '');
datastore.runQuery(q)
```

```
The application will crash due to an uncaught exception:
Error: We attempted to return all of the numeric values, but value 8446744073709552000 is out of bounds of 'Number.MAX_SAFE_INTEGER'.
To prevent this error, please consider passing 'options.wrapNumbers=true' or
'options.wrapNumbers' as
{
  integerTypeCastFunction: provide <your_custom_function>
  properties: optionally specify property name(s) to be cutom casted}
```
Also a typo misspelling of 'custom' here ----^

```
at decodeIntegerValue (/home/jpate/Documents/escaperoom/leaderboard/node_modules/@google-cloud/datastore/build/src/entity.js:388:13)
    at Object.decodeValueProto (/home/jpate/Documents/escaperoom/leaderboard/node_modules/@google-cloud/datastore/build/src/entity.js:462:23)
    at Object.entityFromEntityProto (/home/jpate/Documents/escaperoom/leaderboard/node_modules/@google-cloud/datastore/build/src/entity.js:607:45)
    at Object.decodeValueProto (/home/jpate/Documents/escaperoom/leaderboard/node_modules/@google-cloud/datastore/build/src/entity.js:465:31)
    at value.values.map (/home/jpate/Documents/escaperoom/leaderboard/node_modules/@google-cloud/datastore/build/src/entity.js:446:57)
    at Array.map (<anonymous>)
    at Object.decodeValueProto (/home/jpate/Documents/escaperoom/leaderboard/node_modules/@google-cloud/datastore/build/src/entity.js:446:37)
[...]

PR #591, opened 2020-02-18
feat: add warning when int value is too large during upload

AVaksman:

Towards #588


src/entity.ts
```diff
@@ -580,6 +580,15 @@ export namespace entity {
 
     if (typeof value === 'number') {
       if (value % 1 === 0) {
```
bcoe: what is the purpose of `value % 1`, this seems like it could use a comment (even though it wasn't added by you).

AVaksman: Done.

Would it be preferable to do 
```ts
if (Number.isInteger(value)) {
...
}
```
instead? Here the [method's](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger) name speaks for itself.

bcoe: the method name is nice :+1:

AVaksman: Swapped the logic to`Number.isInteger(value)`

src/entity.ts
```diff
@@ -579,7 +579,16 @@ export namespace entity {
     }
 
     if (typeof value === 'number') {
-      if (value % 1 === 0) {
+      if (Number.isInteger(value)) {
+        if (!Number.isSafeInteger(value)) {
+          process.emitWarning(
+            'Integer value ' +
+              value +
+              ' is outside of bounds of a JavaScript Number\n' +
```
stephenplusplus: Just a thought on some improvements of the messaging:

"Integer value 922337203685477500 is outside of bounds of a JavaScript Number
Use 'Datastore.int(<integer_value_as_string>)' to preserve accuracy during the upload. IntegerOutOfBoundsWarning"

Could be:

"IntegerOutOfBoundsWarning: The value provided is out of bounds of 'Number.MAX_SAFE_INTEGER'.
Use 'Datastore.int(<integer_value_as_string>)' to preserve accuracy during the upload."

We shouldn't include the given integer value in the log, since it won't match what the user actually gave, as the JS engine truncates it:

```js
console.log(9223372036854775807) // 9223372036854776000
```

Is there a way to emit this warning when we also know the property the value comes from, e.g. "The value for `userId` is out of bounds"?

AVaksman: Done 👍 

Changed warning message to 
```
IntegerOutOfBoundsWarning: the value for 'largeInt' property is outside of bounds of a JavaScript Number.
Use 'Datastore.int(<integer_value_as_string>)' to preserve accuracy during the upload.
```

stephenplusplus:
LGTM!

PR #608, opened 2020-03-06
feat: move library to Typescript code generation

xiaozhenliu-gg5:


package.json
```diff
@@ -8,13 +8,12 @@
...
   },
   "repository": "googleapis/nodejs-datastore",
-  "main": "./build/src/index.js",
+  "main": "build/src/index.js",
   "files": [
     "build/proto",
     "build/protos",
     "build/src"
   ],
-  "types": "build/src/index.d.ts",
```
JustinBeckwith: Are you sure we don't need that?

package.json
```diff
@@ -47,49 +47,56 @@
     "prelint": "cd samples; npm link ../; npm i"
   },
   "dependencies": {
-    "@google-cloud/projectify": "^1.0.0",
-    "@google-cloud/promisify": "^1.0.0",
-    "@grpc/grpc-js": "0.6.16",
+    "@google-cloud/projectify": "^1.0.4",
+    "@google-cloud/promisify": "^1.0.4",
+    "@grpc/grpc-js": "0.7.0",
```
JustinBeckwith: @alexander-fenster are we ready for that to sneak in?

synth.py
```diff
@@ -5,24 +5,32 @@
...
 # Copy everything except for top level index, package.json, and README.md
 s.copy(
     library,
-    excludes=['package.json', 'README.md', 'src/index.js'])
-
-# Update path discovery due to build/ dir and TypeScript conversion.
-s.replace("src/v1/datastore_client.js", "../../package.json", "../../../package.json")
+    excludes=['package.json', 'README.md', 'src/index.ts'])
 
+system_test_files=['system-test/fixtures/sample/src/index.ts','system-test/fixtures/sample/src/index.js']
```
JustinBeckwith: I don't think we want this here ...  the system test isn't generate (at least I don't think).

xiaozhenliu-gg5: We do have really simple generated system-test from micro-generator using `packnplay` module, it just creates a new client for ts/js users. And since we have a manual layer for this client library, the client name gets changed from `DatastoreClient` -> `Datastore`. That's why here's the hack :(

package.json
xiaozhenliu-gg5: all tests seem to pass without this, but I'm not sure if we can safely remove it. For my understanding, it's here because we need types for `index.js` and we moved it to `index.ts` so maybe this `index.d.ts` is no longer needed? @alexander-fenster for confirmation, same thing happened to the converting API `bigtable` Thank you!

package.json
xiaozhenliu-gg5: these dependencies are pre-existing, I just moves them around. They might for the samples-test, not sure about that, i can try remove them to see the outcome. Thank you!

synth.py
JustinBeckwith: Yar!  My thinking - maybe we should make the emitting of the system test a feature behind a flag in the gapic generator, then set it to disabled for these bacon wrapped gapics?

alexander-fenster: That's a good idea @JustinBeckwith, even though it's a little bit harder to implement (because of how the template engine works in the generator). I filed https://github.com/googleapis/gapic-generator-typescript/issues/320, we'll see what we can do.

xiaozhenliu-gg5: 👍

package.json
alexander-fenster: We need to use the same version as shipped with google-gax, otherwise it will cause confusion and double installs. Gax uses 0.6.x and will switch to 0.7.x a little bit later, so @xiaozhenliu-gg5 let's use 0.6.x here as well.

package.json
```diff
@@ -47,49 +47,56 @@
...
-    "google-auth-library": "^5.0.0",
-    "google-gax": "^1.7.5",
+    "google-auth-library": "^5.10.1",
+    "google-gax": "^1.14.2",
     "is": "^3.3.0",
     "split-array-stream": "^2.0.0",
     "stream-events": "^1.0.5"
   },
   "devDependencies": {
+    "@types/mocha": "^7.0.2",
```
alexander-fenster: nit: we can sort them

xiaozhenliu-gg5: done! :)

package.json
xiaozhenliu-gg5: Oh, that's what Justin means, move it back to `0.6.x` thanks!

PR #611, opened 2020-03-09
feat: add options param to query#runStream

AVaksman:


stephenplusplus:
@AVaksman could you please explain the motivation of this PR and more about the changes?

AVaksman:
> @AVaksman could you please explain the motivation of this PR and more about the changes?

Sure 
`query.runStream()` delegates to `request.runQueryStream()` and typescript users don't have an ability to pass `options`
```ts
export interface RunQueryOptions {
  consistency?: 'strong' | 'eventual';
  gaxOptions?: CallOptions;
  wrapNumbers?: boolean | IntegerTypeCastOptions;
}
```
in particular the `wrapNumbers` value, leaving the user no way to handle out of bounds integers.

system-test/datastore.ts
```diff
@@ -619,6 +619,32 @@ describe('Datastore', () => {
...
+
+    it('should run a transaction query as a stream via query#runStream', done => {
+      const transaction = datastore.transaction();
+      const q = transaction.createQuery('Character').hasAncestor(ancestor);
+      let resultsReturned = 0;
+      q.runStream()
+        .on('error', done)
+        .on('data', () => resultsReturned++)
+        .on('end', () => {
+          console.log('total results are => ' + resultsReturned);
```
stephenplusplus: I don't think we should log from the tests, just keeping the output clean 👍

stephenplusplus:
Ah, gothca. Thanks! I just had a small change request, but LGTM.

system-test/datastore.ts
AVaksman: Done! 
Thanks

Anything else jumps out?

stephenplusplus: Nope, LGTM. Thanks!

PR #614, opened 2020-03-11
fix: identify propertyName

### LINKED ISSUE #607: Using IntegerTypeCastOptions.properties fails to wrap specified properties
#### Environment details

  - OS: macOS 10.15.2
  - Node.js version: 10.19.0
  - npm version: 6.14.1
  - `@google-cloud/datastore` version: 5.0.6

#### Steps to reproduce

  1. Add an integer larger than `Number.MAX_SAFE_INTEGER` to any Integer property of a datastore entity
  2. Run a query that includes that property in IntegerTypeCastOptions.properties 

#### Details
`_entityPropertyName` is never set in the `Int` constructor because the protoBuf Value object passed to `decodeValueProto()` doesn't have the `propertyName` key that the constructor expects.

Screen Shot 2020-03-06 at 12 59 03 PM

But `Int.valueOf()` only uses the `wrapOptions.integerTypeCastFunction` passed to the query if `this._entityPropertyName` is in the `wrapOptions.properties` Array passed to the query.

https://github.com/googleapis/nodejs-datastore/blob/master/src/entity.ts#L169

Since `this._entityPropertyName` is undefined, `Int.valueOf()` does not custom cast and calls `decodeIntegerValue()`, which throws an Error. Passing `{ wrapOptions: true }` to the query "works", but I don't want *all* the integers on the entity wrapped.

AVaksman:

Fixes #607

### DIFF
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -708,6 +708,7 @@ export namespace entity {
     // tslint:disable-next-line forin
     for (const property in properties) {
       const value = properties[property];
+      value.propertyName = property;
       entityObject[property] = entity.decodeValueProto(value, wrapNumbers);
     }
 
--- a/system-test/datastore.ts
+++ b/system-test/datastore.ts
@@ -257,6 +257,31 @@ describe('Datastore', () => {
       await datastore.delete(postKey);
     });
 
+    it('should wrap specified properties via IntegerTypeCastOptions.properties', async () => {
+      const postKey = datastore.key('Scores');
+      const largeIntValueAsString = '9223372036854775807';
+      const panthers = Datastore.int(largeIntValueAsString);
+      const broncos = 922337203;
+      let integerTypeCastFunctionCalled = 0;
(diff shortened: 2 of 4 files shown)


stephenplusplus:
Sorry, accidental close. @AVaksman could you explain how this change works?

AVaksman:
> @AVaksman could you explain how this change works?
Sure

Originally `decodeValueProto()` only dealt with `valueType` and its `value`. 

With the addition of `IntegerTypeCastOptions` option for `wrapNumbers` [`entity.Int`](https://github.com/googleapis/nodejs-datastore/blob/83d1774cc2ee0e36376d41102f5f71575c467fa0/src/entity.ts#L135-L137) needs the `propertyName` property in case select properties are specified by user in `IntegerTypeCastOptions`.

This PR identifies and adds `propertyName` property to the `valueProto` object which down the line is passed to `entity.Int` constructor.

stephenplusplus:
Thanks!

PR #656, opened 2020-04-15
chore: clean up synth.py

xiaozhenliu-gg5:

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,6 +1,6 @@
 **/node_modules
-**/coverage
-test/fixtures
+**/.coverage
 build/
 docs/
 protos/
+system-test/
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,6 +1,6 @@
 **/node_modules
-**/.coverage
+**/coverage
+test/fixtures
 build/
 docs/
 protos/
-system-test/
--- a/system-test/install.ts
+++ b/system-test/install.ts
@@ -16,34 +16,36 @@
 // ** https://github.com/googleapis/gapic-generator-typescript **
 // ** All changes to this file may be overwritten. **
 
-import { packNTest } from 'pack-n-play';
-import { readFileSync } from 'fs';
-import { describe, it } from 'mocha';
+import {packNTest} from 'pack-n-play';
+import {readFileSync} from 'fs';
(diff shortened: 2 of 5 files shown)


.eslintignore
```diff
@@ -1,6 +1,6 @@
 **/node_modules
-**/coverage
-test/fixtures
+**/.coverage
 build/
 docs/
 protos/
+system-test/
```
JustinBeckwith, line 6: I am a little confused.  Are these hand written changes over top of the generated ignorefile?  This should be overwritten by synth:
https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/node_library/.eslintignore

I think we *want* to lint the `system-test` directory.

system-test/install.ts
```diff
@@ -16,36 +16,34 @@
 // ** https://github.com/googleapis/gapic-generator-typescript **
 // ** All changes to this file may be overwritten. **
 
-import {packNTest} from 'pack-n-play';
-import {readFileSync} from 'fs';
-import {describe, it} from 'mocha';
+import { packNTest } from 'pack-n-play';
```
JustinBeckwith, line 19: Yeah this formatting says it isn't applying `prettier` format rules :)

.eslintignore
xiaozhenliu-gg5: These are all autogeneraed code (I removed the hacks in synth.py about skipping `eslintignore`)
I thought it's not needed any more, so I tried to remove all hacks and see if it works.
I will try upgrade synthtool and run it again, it should be overwritten by synth properly.

xiaozhenliu-gg5:
Some background @JustinBeckwith , I'm iterating all APIs to check the synth.py hacks, see if there are unneeded surgeries in synth script and clean them up. Thanks

system-test/install.ts
xiaozhenliu-gg5: right, double checking. Thanks!

.eslintignore
xiaozhenliu-gg5: Reran synthtool and it's changed back by gts. Thanks!

system-test/install.ts
xiaozhenliu-gg5: This change is back. Thank you!

PR #671, opened 2020-05-18
fix: missing await in rollback

### LINKED ISSUE #668: Transaction rollback is missing await
https://github.com/googleapis/nodejs-datastore/blob/380c45b3a9566037c3daa37d5b5b85ba318fa281/samples/tasks.markdone.js#L47

Transaction rollback is missing await. According to documentation `transaction.rollback()` returns `Promise`.
 
https://googleapis.dev/nodejs/datastore/latest/Transaction.html#rollback

laljikanjareeya:
Fixes #668

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -1304,7 +1304,9 @@ class DatastoreRequest {
         try {
           await transaction.rollback();
         } catch (error) {
-          // rollbacks are fine as best effort.
+          // Provide the error & API response from the failed run to the user.
+          // Even a failed rollback should be transparent.
+          // RE: https://github.com/GoogleCloudPlatform/gcloud-node/pull/1369#discussion_r66833976
         }
         callback!(err);
         return;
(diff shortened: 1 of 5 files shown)


src/request.ts
```diff
@@ -1321,7 +1325,7 @@ class DatastoreRequest {
         const [response] = await transaction.commit();
         callback!(null, response);
       } catch (err) {
-        transaction.rollback();
+        await transaction.rollback();
```
AVaksman, line 1328: should this also be wrapped in `try-catch` as above?

src/request.ts
```diff
@@ -1301,7 +1301,11 @@ class DatastoreRequest {
     const transaction = this.datastore.transaction();
     transaction.run(async err => {
       if (err) {
-        transaction.rollback();
+        try {
+          await transaction.rollback();
+        } catch (error) {
+          // rollbacks are fine as best effort.
```
AVaksman, line 1307: Can probably add reference to a related discussion
```ts
// Even a failed rollback should be transparent.
// RE: https://github.com/GoogleCloudPlatform/gcloud-node/pull/1369#discussion_r66833976
```

laljikanjareeya:
@AVaksman  It shows code coverage decrease in `src/index.ts` but, there is not change on that file.

laljikanjareeya:
@AVaksman Addressed all the comments please take a look!!

PR #694, opened 2020-06-29
fix!: transaction.merge

### LINKED ISSUE #547: Transaction does not support with merge and have unexpected promise error out of the process
Below codes will prompt " UnhandledPromiseRejectionWarning: Error: 10 ABORTED: too much contention on these datastore entities. please try again." without able to be caught in any try catch block.

If I use transaction.save() or transaction.update(), things work normally.

===============

```typescript
async function merge(taskKey: any) {
    const transaction = datastore.transaction();
    try {
        await transaction.run();
        const [myTask] = await transaction.get(taskKey);
        const newTask = {
            key: taskKey,
            data: {
                value: 1,
            },
        };
        transaction.merge(newTask);
        await transaction.commit();
    } catch (err) {
        console.log("error", err.message);
        await transaction.rollback();
        throw err;
    }
}

async function test3() {
    // The Cloud Datastore key for the new entity
    const kind = "transaction";
    const taskKey = datastore.key([kind, 1]);

    // Prepares the new entity
    const task = {
        key: taskKey,
        data: {
            value: 10,
        },
    };
    await datastore.save(task);

    // Saves the entity
    try {
        const promises = [
            merge(taskKey),
            merge(taskKey),
        ];
        await Promise.all(promises);
    } catch (err) {
        console.log("error", err.message);
    }
}

```

AVaksman:
BREAKING CHANGE: transaction.merge(): is now void and accepts two single `entity` objects to merge as params.

Fixes #547


src/transaction.ts
```diff
@@ -677,6 +677,50 @@ class Transaction extends DatastoreRequest {
...
+   *   });
+   * });
+   */
+  merge(currentEntity: Entity, entityToMerge: Entity): void {
+    const preparedCurrent = Datastore.prepareEntityObject_(currentEntity);
+    const preparedToMerge = Datastore.prepareEntityObject_(entityToMerge);
+    Object.assign(
+      preparedCurrent.data,
+      preparedToMerge.data,
+      preparedCurrent.data
```
stephenplusplus, line 719: I noticed that we go essentially "Object.assign(a, b, a)" -- what's the ordering significance here?

stephenplusplus:
Could you explain more about why this will use different signatures for "datastore.merge" and "transaction.merge"?

JustinBeckwith (changes requested):
We just shipped a 6.0 in June.  I would prefer to avoid a breaking change at this stage.

AVaksman:
> Could you explain more about why this will use different signatures for "datastore.merge" and "transaction.merge"?

Sure.
`datastore.merge` accepts one or more objects where API would
for each Entity object it will
1. Begin a transaction
2. For each object provided by user API would retrieve an object from server by key
3. Merge the data from both objects
4. Set method to `upsert`
5. Add the merged object to `modifiedEntities_` array (`transaction.save`)
6. Once finished iterating through all user provided objects, finish off with `transaction.commit`.

`transaction.merge` is void and needs two objects to do 3, 4 and 5 from the above list.

src/transaction.ts
AVaksman: Fixed, thanks!

stephenplusplus:
Thanks. Even inside of a transaction, a user could want to merge multiple entities too though, right?

src/index.ts
```diff
@@ -811,6 +811,70 @@ class Datastore extends DatastoreRequest {
...
+   * @param {function} callback The callback function.
+   * @param {?error} callback.err An error returned while making this request
+   * @param {object} callback.apiResponse The full API response.
+   *
+   * @example
+   * const {Datastore} = require('@google-cloud/datastore');
+   * const datastore = new Datastore();
+   *
+   * const key = datastore.key(['Company', 123]);
+   * datastore.merge({meaningOfLife: 42}, (err) => {
```
stephenplusplus, line 844: The key doesn’t seem to be used inside of an Entity here.

AVaksman: Updated, thanks

AVaksman:
> Thanks. Even inside of a transaction, a user could want to merge multiple entities too though, right?

Yes, at least two entities are needed for `transaction.merge`

stephenplusplus (approved):

JustinBeckwith (approved):

JustinBeckwith:
@AVaksman just to double check - this now doesn't have a breaking change, right?  Are we good to merge?

AVaksman:
I believe this PR still has a breaking change.
`Transaction.merge()` signature is being changed from 
```ts
merge(entities: Entity | Entity[])
```
to 
```ts
merge(currentEntity: Entity, entityToMerge: Entity): void
```

bcoe:
we should consider dusting this off when we drop Node 10, but I think for now we can close.

src/transaction.ts
```diff
@@ -402,6 +402,46 @@ class Transaction extends DatastoreRequest {
...
+   *
+   *     transaction.commit((err) => {
+   *       if (!err) {
+   *         // Data saved successfully.
+   *       }
+   *     });
+   *   });
+   * });
+   */
+  merge(currentEntity: Entity, entityToMerge: Entity): void {
```
danieljbruce, line 437: I see that now we change this signature to allow for the flexibility of fetching entities beforehand in a separate transaction/operation and then providing them and using save because multiple save operations can be run at the same time without an error. But is the breaking change really worth it?

It will mean:
- Users need to change how they use merge
- They can currently work around this merge limitation by using `.save` anyway
- Not sure yet, but there may be a chance that just deleting local merging at https://github.com/googleapis/nodejs-datastore/blob/d854b57ed2b9b9f9ea148daedeb6d64e4d99de4d/src/request.ts#L956 gives us what we need

danieljbruce: I take back the third point. I hadn't realized `upsert` is different from a `merge` in that for a particular key, `upsert` replaces the current data entirely and merge merges the properties of the entity passed in with the entity currently saved.

I think we could add a new option for such merges to be handled by the backend and then we do not need a breaking change.

PR #758, opened 2020-11-18
refactor: a couple of minor refactoring.

vicb:


vicb:
> Could you describe the change being made? I may be being a bit dense here and I can't quite tease it out.

Sure

"refactor: take advantage of TS type inference"

TS infers the type to `SomeType` if the function return a boolean = true as `is SomeType`.

```typescript
function isSomeType(value: any): is SomeType {
  return value instanceof SomeType;
}

if (isSomeType(someVar) {
 // someVar to be an instance of SomeType
}
```

"refactor: minor code simplification"

Drop a while use in favor or `String.prototype.repeat()` - available since [Node 4](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat#Browser_compatibility).

bcoe (approved):

Issue #755, opened 2020-11-17
Saving an empty Buffer throws an error.


vicb:
#### Environment details

  - OS: MacOs
  - Node.js version: v14.4.0
  - npm version: 6.14.8
  - `@google-cloud/datastore` version: 6.3.0

#### Steps to reproduce

Execute the following code:

```typescript
async function saveBlob() {
  const key = datastore.key(['TEST']);
  
  await datastore.save({
    key: key,
    data: {
      name: 'test',
      blob: Buffer.from([])
    },
  });
}
```

It will trigger an error: `Error: 3 INVALID_ARGUMENT: The value "blob" does not contain a value.`

I expect to be able to save an empty Buffer without error.

If you change the blob value to `Buffer.from([1, 2, 3])` the error disappears.

My use case is that I have a protocol buffer composed of repeated fields only:

```proto
message LiveTrack {
  optional int64 id = 1 [jstype = JS_NUMBER];
  optional string name = 2;   
  repeated float lat = 3;
  repeated float lon = 4;
  repeated int32 alt = 5;
  repeated int32 time_sec = 6;
  repeated uint32 flags = 7;
  map<uint32, LiveExtra> extra = 8;
}
```

It could happen that all repeated fields (including the map) are empty. Because the optional field are not populated, the proto would serialize to an empty `Buffer`.

**Workaround**

The workaround would be to set the value to `null` when the Buffer size is 0.
This requires extra code both while saving (to replace the `Buffer` with `null`) and while loading (to replace `null` with the default message). 

**Other**

The wording of the message could be improved: `The value "blob" does not contain a value.`, i.e. use field or key instead of the first "value".

vicb:
I have update `encodeValue` to read like:

```typescript
        if (value instanceof Buffer) {
            valueProto.blobValue = value.length == 0 ? '' : buffer;
            return valueProto;
        }
```

And it fixes the issue:
- The column is save as a Blob without vale (verified in the Cloud Console),
- Reading the column returns an empty buffer.

I suspect that the issue might originate in the library used to encode the protocol buffer ?

Anyway this code looks like a fix.

What do you think ?

vicb:
I have done some digging into this bug:

/Users/berchet/Code/datastore/node_modules/@grpc/proto-loader/build/src/index.js

[...]

PR #767, opened 2020-11-21
fix: storing an empty Buffer

vicb:

As discussed in #755 datastore can not currently store an empty buffer due to a bug in protobufjs - see protobufjs/protobuf.js#1500.

However protobufs supports 2 way of passing a buffer, as seen in [converter.js](https://github.com/protobufjs/protobuf.js/blob/aea41b87da59939b51b5bab89cc0180c8838f8ab/src/converter.js#L71-L75):

```javascript
            case "bytes": gen
                ("if(typeof d%s===\"string\")", prop)
                    ("util.base64.decode(d%s,m%s=util.newBuffer(util.base64.length(d%s)),0)", prop, prop, prop)
                ("else if(d%s.length)", prop)
                    ("m%s=d%s", prop, prop);
```

This PR switch from passing a `Buffer` (the else branch in the snippet above) to passing a base64 encoded string (the if branch in the snippet above).

Encoding/Decoding a non empty buffer was already tested and I added a test for encoding/decoding an empty buffer. As expected the test fails with the current code:

```
  1) Datastore
       create, retrieve and delete
         should save/get/delete an empty buffer:
     Error: 3 INVALID_ARGUMENT: The value "buf" does not contain a value.
```

I am not sure if #755 should be marked as fixed or if we should wait for protobufs to be fixed ?

### DIFF
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -629,7 +629,10 @@ export namespace entity {
     }
 
     if (value instanceof Buffer) {
-      valueProto.blobValue = value;
+      // Convert the buffer to a base 64 string to workaround a bug of
+      // protobufs encoding empty buffer.
+      // See https://github.com/googleapis/nodejs-datastore/issues/755
+      valueProto.blobValue = value.toString('base64');
       return valueProto;
     }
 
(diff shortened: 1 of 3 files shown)


crwilcox (approved):

crwilcox:
Thanks for the contribution!

vicb:
Thanks for merging the PR Chris.

Issue #760, opened 2020-11-18
Remove Number to Datastore Type inference from save and query.


crwilcox:
Related to #754, #759 

Remove auto inference of Numbers. This is a 'foot-gun' in the library and we should encourage usage of `datastore.double` and `datastore.int` so the library doesn't choose poorly.

This is for sure used in Save and Query today.

vicb:
It would be nice to create a migration path for DB created before inference is removed.

In such DB there could be a mix of integer/float for the same property and `datastore.double` / `datastore.int` would only apply to numbers created after the inference is removed.

crwilcox:
Going to block on this happening for a bit. Warning may be enough and is in PR, tracked by 759.

crwilcox:
I am rewording this as the approach has moved from removal to changing the default behavior. By default we shouldn't unwrap these. Int and Double , as of the pr (773) extend Number. Keeping them wrapped should cause no/minimal friction, but allow us to still have additional data about the type in datastore.

meredithslota:
Reconsider the approach (back to removal) — added the`next major: breaking change` label to ensure we don't miss it.

Issue #754, opened 2020-11-12
[Bug] Query with filter on an integer column.


vicb:
1) Is this a client library issue or a product issue?

I guess a bug in the client library (`"@google-cloud/datastore": "^6.3.0"`)

#### Environment details

  - OS: Linux
  - Node.js version: 14.15 
  - npm version: 6.14.8
  - `@google-cloud/datastore` version: 6.3.0

#### Steps to reproduce

1) Create an entity in the cloud console with an integer column, named `integer_column` and set its value to 200.

2) Executes a query with a filter on that column:

```typescript
    const query = datastore
      .createQuery(tableName)
      .filter('integer_column', '>', value);

    const [results] = await datastore.runQuery(query);
```

**Expected behavior:**

The correct results are returned if `value` is a valid numerical value, i.e. 123 or 123.456

**Actual behavior**

The correct results are only returned if value is 123. A float value (123.456) would return an empty result.

**Workaround**

Update the query to add `Math.round()`.

```typescript
    const query = datastore
      .createQuery(tableName)
      .filter('integer_column', '>', Math.round(value));

    const [results] = await datastore.runQuery(query);
```

Thanks!

stephenplusplus:
Thanks for opening the issue! What would be an ideal solution on our side? Running `Math.floor()` on the float? Throw an error, specifying only an integer is expected?

vicb:
The ideal solution is that `.filter('integer_column', '>', value)` works whatever the (numerical) value.

I mean `10 > 5.123 == true` is something everybody expects to work, whatever the language/API.

stephenplusplus:
The issue I'm seeing is that the API expects an integer value to compare against integers. It cannot compare floats against integers. So, if we're given a float of `123.456`, we would have to lose precision to reduce it to `123`. Of course, the results could never include a result where `integer_column` contains `123.456`, so a result with `124` will come through in both cases, as that's the next possible value. What I'd be concerned about, is that we'd be adding code that could be misleading-- having the user assume it's possible to have float precision where they cannot. It may be best to continue sanitizing non-integer input in the application layer, to prepare it for the integer operation.

vicb:
1) Again everybody expects `10 > 5.123 == true`. The opposite is **very** misleading.

2) Thoughts:
 
> So, if we're given a float of `123.456`, we would have to lose precision to reduce it to `123`

You have to be careful about the rounding here:
[...]

PR #773, opened 2020-12-08
fix!: don't unwrap int, double by default, add a warning when a number is autoconverted to double or int

### LINKED ISSUE #759: Add warning when inference used for Number types to Datastore types
As discussed in #754, the auto conversion to datastore types doesn't work particualrly well in Node.js. Numbers need to be coverted to int, double, but there are times it makes the wrong choice.

This functionality is planned to be removed in a future version as it is too easy for customers to do the wrong thing and end up with disparate types unexpectedly.

Example
Insert "3.0", "3.99", "4.99" expecting double, double, double. Actual is int, double, double, as 3.0 is an integer.

This issue tracks adding a warning in a fix release letting customers know that inference is being used and can be fragile, particularly if used for columns expected to be floats/doubles.

### LINKED ISSUE #7165: Add warning when inference used for Number types to Datastore types
As discussed in googleapis/google-cloud-node#7166, the auto conversion to datastore types doesn't work particualrly well in Node.js. Numbers need to be coverted to int, double, but there are times it makes the wrong choice.

This functionality is planned to be removed in a future version as it is too easy for customers to do the wrong thing and end up with disparate types unexpectedly.

Example
Insert "3.0", "3.99", "4.99" expecting double, double, double. Actual is int, double, double, as 3.0 is an integer.

This issue tracks adding a warning in a fix release letting customers know that inference is being used and can be fragile, particularly if used for columns expected to be floats/doubles.

crwilcox:
BREAKING CHANGE: Datastore doubles and ints are no longer automatically decoded to Javascript Numbers
BREAKING CHANGE: a warning is emitted whenever autoconversion from a Javascript Number to a Datastore Double or Int is used.

As the autoconversion from number to int/double can result in inconsistent db representations, this warns to not use this and that there is some danger to doing so.

Take the double value 4.0. While in Node.js this is consider to be an int, a user specifying 4.0 likely intends for that to be a Datastore Double. The automatic conversion can result in mixed type columns which can make querying entities and their ordering odd, as at this time Datastore entities are ordered by type, then value.

Fixes googleapis/google-cloud-node#7165, googleapis/google-cloud-node#7162, googleapis/google-cloud-node#7164 🦕

### DIFF
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -52,27 +52,46 @@ export namespace entity {
    * provided.
    *
    * @class
-   * @param {number} value The double value.
+   * @param {number|string} value The double value.
    *
    * @example
    * const {Datastore} = require('@google-cloud/datastore');
    * const datastore = new Datastore();
    * const aDouble = datastore.double(7.3);
    */
-  export class Double {
+  export class Double extends Number {
+    private _entityPropertyName: string | undefined;
     type: string;
-    value: number;
-    constructor(value: number) {
+    value: string;
+    constructor(value: number | string | ValueProto) {
+      super(typeof value === 'object' ? value.doubleValue : value);
+
       /**
        * @name Double#type
        * @type {string}
        */
       this.type = 'DatastoreDouble';
+
+      this._entityPropertyName =
+        typeof value === 'object' ? value.propertyName : undefined;
+
       /**
        * @name Double#value
-       * @type {number}
+       * @type {string}
        */
-      this.value = value;
+      this.value =
+        typeof value === 'object'
+          ? value.doubleValue.toString()
+          : value.toString();
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    valueOf(): any {
+      return Number(this.value);
+    }
+
+    toJSON(): Json {
+      return {type: this.type, value: this.value};
     }
   }
 
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -16,12 +16,11 @@
 import arrify = require('arrify');
 import * as extend from 'extend';
 import * as is from 'is';
-import {Query, QueryProto, IntegerTypeCastOptions} from './query';
-import {PathType} from '.';
-import {protobuf as Protobuf} from 'google-gax';
+import { Query, QueryProto, IntegerTypeCastOptions } from './query';
+import { PathType } from '.';
+import { protobuf as Protobuf } from 'google-gax';
 import * as path from 'path';
-import {google} from '../protos/protos';
-import {type} from 'os';
+import { google } from '../protos/protos';
 
 // eslint-disable-next-line @typescript-eslint/no-namespace
 export namespace entity {
@@ -34,7 +33,7 @@ export namespace entity {
       const errorMessages = {
         MISSING_KIND: 'A key should contain at least a kind.',
         MISSING_ANCESTOR_ID: 'Ancestor keys require an id or name.',
-      } as {[index: string]: string};
+      } as { [index: string]: string };
       super(errorMessages[opts.code]);
       this.name = 'InvalidKey';
     }
@@ -189,7 +188,7 @@ export namespace entity {
     }
 
     toJSON(): Json {
-      return {type: this.type, value: this.value};
+      return { type: this.type, value: this.value };
     }
   }
 
@@ -447,16 +446,16 @@ export namespace entity {
     if (!Number.isSafeInteger(num)) {
       throw new Error(
         'We attempted to return all of the numeric values, but ' +
-          (value.propertyName ? value.propertyName + ' ' : '') +
-          'value ' +
-          value.integerValue +
-          " is out of bounds of 'Number.MAX_SAFE_INTEGER'.\n" +
-          "To prevent this error, please consider passing 'options.wrapNumbers=true' or\n" +
-          "'options.wrapNumbers' as\n" +
-          '{\n' +
-          '  integerTypeCastFunction: provide <your_custom_function>\n' +
-          '  properties: optionally specify property name(s) to be custom casted\n' +
-          '}\n'
+        (value.propertyName ? value.propertyName + ' ' : '') +
+        'value ' +
+        value.integerValue +
+        " is out of bounds of 'Number.MAX_SAFE_INTEGER'.\n" +
+        "To prevent this error, please consider passing 'options.wrapNumbers=true' or\n" +
+        "'options.wrapNumbers' as\n" +
+        '{\n' +
+        '  integerTypeCastFunction: provide <your_custom_function>\n' +
+        '  properties: optionally specify property name(s) to be custom casted\n' +
+        '}\n'
       );
     }
     return num;
@@ -611,13 +610,13 @@ export namespace entity {
         if (!Number.isSafeInteger(value)) {
           process.emitWarning(integerOutOfBoundsWarning);
         } else {
-          process.emitWarning(typeCastWarning);
+          warn('TypeCastWarning', typeCastWarning);
         }
         value = new entity.Int(value);
         valueProto.integerValue = value.value;
         return valueProto;
       } else {
-        process.emitWarning(typeCastWarning);
+        warn('TypeCastWarning', typeCastWarning);
         value = new entity.Double(value);
         valueProto.doubleValue = value.value;
         return valueProto;
@@ -682,6 +681,15 @@ export namespace entity {
     }
 
     throw new Error('Unsupported field value, ' + value + ', was provided.');
+  }
+
+  const warningTypesIssued = new Set<string>();
+  const warn = (warningName: string, warningMessage: string) => {
+    console.log(warningName);
+    if (!warningTypesIssued.has(warningName)) {
+      warningTypesIssued.add(warningName);
+      process.emitWarning(warningMessage);
+    }
   }
 
   /**
@@ -1334,7 +1342,7 @@ export namespace entity {
       const reference = {
         app: projectId,
         namespace: key.namespace,
-        path: {element: elements},
+        path: { element: elements },
       };
 
       const buffer = this.protos.Reference.encode(reference).finish();
@@ -1431,7 +1439,7 @@ export interface ValueProto {
 
 export interface EntityProto {
   key?: KeyProto | null;
-  properties?: {[k: string]: ValueProto};
+  properties?: { [k: string]: ValueProto };
   excludeFromIndexes?: boolean;
 }
 
@@ -1455,7 +1463,7 @@ export interface ResponseResult {
 }
 
 export interface EntityObject {
-  data: {[k: string]: Entity};
+  data: { [k: string]: Entity };
   excludeFromIndexes: string[];
 }
 
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -475,7 +475,7 @@ export namespace entity {
    *
    * @private
    * @param {object} valueProto The protobuf Value message to convert.
-   * @param {boolean | IntegerTypeCastOptions} [wrapNumbers=false] Wrap values of integerValue type in
+   * @param {boolean | IntegerTypeCastOptions} [wrapNumbers=true] Wrap values of integerValue type in
    *     {@link Datastore#Int} objects.
    *     If a `boolean`, this will wrap values in {@link Datastore#Int} objects.
    *     If an `object`, this will return a value returned by
@@ -523,10 +523,21 @@ export namespace entity {
       }
 
       case 'doubleValue': {
+        if (wrapNumbers === undefined) {
+          wrapNumbers = true;
+        }
+
+        if (wrapNumbers) {
+          return new entity.Double(value);
+        }
         return Number(value);
       }
 
       case 'integerValue': {
+        if (wrapNumbers === undefined) {
+          wrapNumbers = true;
+        }
+
         return wrapNumbers
           ? typeof wrapNumbers === 'object'
             ? new entity.Int(valueProto, wrapNumbers).valueOf()
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -547,7 +547,7 @@ export namespace entity {
         }
 
         if (wrapNumbers) {
-          return new entity.Double(value);
+          return new entity.Double(valueProto);
         }
         return Number(value);
       }
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -596,7 +596,8 @@ export namespace entity {
         "'IntegerOutOfBoundsWarning: the value for '" +
         property +
         "' property is outside of bounds of a JavaScript Number.\n" +
-        "Use 'Datastore.int(<integer_value_as_string>)' to preserve accuracy during the upload.";
+        "Use 'Datastore.int(<integer_value_as_string>)' to preserve accuracy " +
+        'in your database.';
 
       const typeCastWarning =
         "TypeCastWarning: the value for '" +
@@ -604,7 +605,7 @@ export namespace entity {
         "' property is a JavaScript Number.\n" +
         "Use 'Datastore.int(<integer_value_as_string>)' or " +
         "'Datastore.double(<double_value_as_string>)' to preserve consistent " +
-        'Datastore types during the upload.';
+        'Datastore types in your database.';
 
       if (Number.isInteger(value)) {
         if (!Number.isSafeInteger(value)) {
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -685,7 +685,6 @@ export namespace entity {
 
   const warningTypesIssued = new Set<string>();
   const warn = (warningName: string, warningMessage: string) => {
-    console.log(warningName);
     if (!warningTypesIssued.has(warningName)) {
       warningTypesIssued.add(warningName);
       process.emitWarning(warningMessage);
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -593,7 +593,7 @@ export namespace entity {
 
     if (typeof value === 'number') {
       const integerOutOfBoundsWarning =
-        "'IntegerOutOfBoundsWarning: the value for '" +
+        "IntegerOutOfBoundsWarning: the value for '" +
         property +
         "' property is outside of bounds of a JavaScript Number.\n" +
         "Use 'Datastore.int(<integer_value_as_string>)' to preserve accuracy " +
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -16,11 +16,11 @@
 import arrify = require('arrify');
 import * as extend from 'extend';
 import * as is from 'is';
-import {Query, QueryProto, IntegerTypeCastOptions} from './query';
-import {PathType} from '.';
-import {protobuf as Protobuf} from 'google-gax';
+import { Query, QueryProto, IntegerTypeCastOptions } from './query';
+import { PathType } from '.';
+import { protobuf as Protobuf } from 'google-gax';
 import * as path from 'path';
-import {google} from '../protos/protos';
+import { google } from '../protos/protos';
 
 // eslint-disable-next-line @typescript-eslint/no-namespace
 export namespace entity {
@@ -33,7 +33,7 @@ export namespace entity {
       const errorMessages = {
         MISSING_KIND: 'A key should contain at least a kind.',
         MISSING_ANCESTOR_ID: 'Ancestor keys require an id or name.',
-      } as {[index: string]: string};
+      } as { [index: string]: string };
       super(errorMessages[opts.code]);
       this.name = 'InvalidKey';
     }
@@ -188,7 +188,7 @@ export namespace entity {
     }
 
     toJSON(): Json {
-      return {type: this.type, value: this.value};
+      return { type: this.type, value: this.value };
     }
   }
 
@@ -216,7 +216,7 @@ export namespace entity {
     return (
       isDsInt(maybeDsInt) ||
       (is.object(maybeDsInt) &&
-        is.string(maybeDsInt.value) &&
+        (is.string(maybeDsInt.value) || is.number(maybeDsInt.value)) &&
         maybeDsInt.type === 'DatastoreInt')
     );
   }
@@ -446,16 +446,16 @@ export namespace entity {
     if (!Number.isSafeInteger(num)) {
       throw new Error(
         'We attempted to return all of the numeric values, but ' +
-          (value.propertyName ? value.propertyName + ' ' : '') +
-          'value ' +
-          value.integerValue +
-          " is out of bounds of 'Number.MAX_SAFE_INTEGER'.\n" +
-          "To prevent this error, please consider passing 'options.wrapNumbers=true' or\n" +
-          "'options.wrapNumbers' as\n" +
-          '{\n' +
-          '  integerTypeCastFunction: provide <your_custom_function>\n' +
-          '  properties: optionally specify property name(s) to be custom casted\n' +
-          '}\n'
+        (value.propertyName ? value.propertyName + ' ' : '') +
+        'value ' +
+        value.integerValue +
+        " is out of bounds of 'Number.MAX_SAFE_INTEGER'.\n" +
+        "To prevent this error, please consider passing 'options.wrapNumbers=true' or\n" +
+        "'options.wrapNumbers' as\n" +
+        '{\n' +
+        '  integerTypeCastFunction: provide <your_custom_function>\n' +
+        '  properties: optionally specify property name(s) to be custom casted\n' +
+        '}\n'
       );
     }
     return num;
@@ -580,29 +580,12 @@ export namespace entity {
       return valueProto;
     }
 
-    if (typeof value === 'number') {
-      if (Number.isInteger(value)) {
-        if (!Number.isSafeInteger(value)) {
-          process.emitWarning(
-            'IntegerOutOfBoundsWarning: ' +
-              "the value for '" +
-              property +
-              "' property is outside of bounds of a JavaScript Number.\n" +
-              "Use 'Datastore.int(<integer_value_as_string>)' to preserve accuracy during the upload."
-          );
-        }
-        value = new entity.Int(value);
-      } else {
-        value = new entity.Double(value);
-      }
-    }
-
-    if (isDsInt(value)) {
+    if (isDsInt(value) || isDsIntLike(value)) {
       valueProto.integerValue = value.value;
       return valueProto;
     }
 
-    if (isDsDouble(value)) {
+    if (isDsDouble(value) || isDsDoubleLike(value)) {
       valueProto.doubleValue = value.value;
       return valueProto;
     }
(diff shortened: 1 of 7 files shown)


src/entity.ts
```diff
@@ -607,6 +591,38 @@ export namespace entity {
...
+        "' property is outside of bounds of a JavaScript Number.\n" +
+        "Use 'Datastore.int(<integer_value_as_string>)' to preserve accuracy during the upload.";
+
+      const typeCastWarning =
+        "TypeCastWarning: the value for '" +
+        property +
+        "' property is a JavaScript Number.\n" +
+        "Use 'Datastore.int(<integer_value_as_string>)' or " +
+        "'Datastore.double(<double_value_as_string>)' to preserve consistent " +
+        'Datastore types during the upload.';
```
vicb, line 607: I would suggest to change "during the upload." for "in your database".

Also it would be good to have some documentation to refer to.

src/entity.ts
```diff
@@ -607,6 +591,38 @@ export namespace entity {
...
+
+      if (Number.isInteger(value)) {
+        if (!Number.isSafeInteger(value)) {
+          process.emitWarning(integerOutOfBoundsWarning);
+        } else {
```
vicb, line 612: Do we need the else here ? 
I don't think the branches are exclusive.

Also maybe move `process.emitWarning(typeCastWarning);` right after l. 594 as soon as we know that the type is a number ?

src/entity.ts
```diff
@@ -607,6 +591,38 @@ export namespace entity {
...
+          process.emitWarning(typeCastWarning);
+        }
+        value = new entity.Int(value);
+        valueProto.integerValue = value.value;
+        return valueProto;
+      } else {
+        process.emitWarning(typeCastWarning);
```
vicb, line 619: As explained in my previous comment if would be nice to factor that up ?

vicb:
@crwilcox I have added a few inline comments.

A few questions:
1) Could you please confirm that this code is called from **both** query and save ?
2) What about an option to silence the "Number" warning ? The rationale would be that the warning will start trigger as soon as users save (and query ?) an integer value with the updated code. I think giving them an option to silence the warning while they update their code would be nice ?

src/entity.ts
crwilcox: They are exclusive. They both recommend the same outcome, but I think having two separate warnings is a bit confusing. This one warning is it could be dangerous, the other is destructive. I had what you recommended earlier but it is very very noisy :)

crwilcox:
> * Could you please confirm that this code is called from **both** query and save ?

Correct. 

> What about an option to silence the "Number" warning ? The rationale would be that the warning will start trigger as soon as users save (and query ?) an integer value with the updated code. I think giving them an option to silence the warning while they update their code would be nice ?

I don't think I am in favor of allowing silencing a warning. The warning exists because this is unsafe. This is already the softer approach to discontinuing this altogether (see the first commit).

src/entity.ts
crwilcox: Sure. I will update the other warning to match (this was phrased this way to be consistent)

src/entity.ts
crwilcox: Resolving as the splitting is only due to the other discussion, would be refactored if we went to raising two warning simultaneously.

vicb:
> I don't think I am in favor of allowing silencing a warning. The warning exists because this is unsafe. This is already the softer approach to discontinuing this altogether (see the first commit).

`Datastore.UNSAFE_CODE_MUST_BE_UPDATED_silence_number_warning = true` ?

test/entity.ts
```diff
@@ -744,13 +744,26 @@ describe('entity', () => {
       assert.deepStrictEqual(entity.encodeValue(value), expectedValueProto);
     });
 
-    it('should encode an int', () => {
+    it.only('should encode an int', done => {
```
JustinBeckwith, line 747: booop

JustinBeckwith (changes requested):

vicb:
I was wondering about the meaning of "!" in the commit message type.

Just looked that up. "!" is for breaking change.

I don't think it applies here ?

vicb:
What about:
1. Updating #759 with a rationale about why it has been re-opened ?
2. Not merging this PR until there is migration path for old DB as discussed in #760 ?
3. Not merging this PR until there is some docs detailing the issue ?

More context about 2: if users fix their code now, they will no get any warning any more but they DB might still be in a "bad state" until they fix it using *a solution to be provided*.

src/entity.ts
crwilcox: Note: added 'debouncing' you will only get a single typecast warning for the entity. that way it doesn't flood the warning feed.

vicb:
Thinking more about this PR, I really think it would be premature to merge the code as is.

1) Inconsistent API

Imagine the case:

```typescript
// Retrieve an entity with *any* of the ds methods.
const [entity] = await ds.get(id);

// [...]

// This would warn because the numbers are not wrapped.
await ds.save(entity);
```

2) No migration path

3) No doc explaining the issue

A more logical way to move forward would be:
1) Document the issue,
2) Propose/Discuss a few different solutions,
3) Implement whatever seems a solid and future proof solution.

To start with may be add some comments in the PR description about when this is intended to be merged ?

src/entity.ts
feywind: It might be reasonable to augment the out of bounds warning to mention the type issue as well, since that'll still be an issue if they fix their number size issue.

src/entity.ts
```diff
@@ -667,6 +683,15 @@ export namespace entity {
     throw new Error('Unsupported field value, ' + value + ', was provided.');
   }
 
+  const warningTypesIssued = new Set<string>();
+  const warn = (warningName: string, warningMessage: string) => {
+    console.log(warningName);
```
feywind, line 688: It feels a bit odd to me to both `console.log` and `process.emitWarning`, as the default warning handler prints the error too. If we're keeping both, `console.warn` seems like the thing to do here. (Or is this a debug line that was left in?)

test/entity.ts
```diff
@@ -744,32 +744,77 @@ describe('entity', () => {
...
+        "' property is a JavaScript Number.\n" +
+        "Use 'Datastore.int(<integer_value_as_string>)' or " +
+        "'Datastore.double(<double_value_as_string>)' to preserve consistent " +
+        'Datastore types in your database.';
+      process.once('warning', warning => {
+        assert.strictEqual(warning.message, expectedWarning);
+        done();
+      });
+
+      entity.Int = function (value_: {}) {
```
feywind, line 767: The monkey-patch here weirds me a bit. If Node is doing any kind of module caching, then the `require()` in `beforeEach()` might not give you a new copy each time. (Which might be the source of the test failures.) It might be safer to use Sinon stubs.

test/entity.ts
```diff
@@ -744,32 +744,77 @@ describe('entity', () => {
...
+        "' property is a JavaScript Number.\n" +
+        "Use 'Datastore.int(<integer_value_as_string>)' or " +
+        "'Datastore.double(<double_value_as_string>)' to preserve consistent " +
+        'Datastore types in your database.';
+      process.once('warning', warning => {
+        assert.strictEqual(warning.message, expectedWarning);
+        done();
+      });
+
       entity.Int = function (value_: {}) {
```
feywind, line 795: Ah, I see, this was already in there. Anyhow, I think my earlier comment is still meaningful to check.

test/entity.ts
```diff
@@ -744,32 +744,77 @@ describe('entity', () => {
...
         assert.strictEqual(value_, value);
         this.value = value_;
       };
 
       assert.deepStrictEqual(entity.encodeValue(value), expectedValueProto);
+      // if an error is reported on this, done() is called more than once and
+      // should cause failure.
+      assert.deepStrictEqual(entity.encodeValue(value), expectedValueProto);
```
feywind, line 803: Looks like this line got doubled.

src/entity.ts
crwilcox: 'doh'

test/entity.ts
crwilcox: This one is intentional. By calling this twice, we ensure the warning is 'debounced'. If it isn't the second call will call done again and be reported as an error

test/entity.ts
crwilcox: I think this is why we delete it from the cache?

feywind: Oh geez, I somehow missed that.

crwilcox:
@vicb, I think most of your comments are handled well enough. I have no intent to write a utility to clean up mixed states across entities. I expect that most customers haven't even encountered this issue as it requires having 'int-able' doubles. That said, you raise another thing I think does require further discussion. Warning on saving a gotten entity.

As authored, if you have a double value that is intable. Also the way this all works with wrapNumbers. If you use ints, you can wrapNumbers, but at this point, if it is a DatastoreDouble wrapNumbers has no effect. 

I am going to commit the following system test that will pass because the test captures current behavior, not necessarily desired behavior.


```
 it('should save/get an int-able double value via Datastore.', async () => {
      const postKey = datastore.key('Team');
      const points = Datastore.double(2);
      await datastore.save({ key: postKey, data: { points } });
      var [entity] = await datastore.get(postKey, { wrapNumbers: true });
      // Expect content is stored in datastore as a double, returned as int
      // as doubles can safely be brought to local types.
      // NOTE: this is a bit awkward as is since you can't prevent decoding of doubles
      // assert.strictEqual(entity.points.type, 'DatastoreDouble');
      // assert.strictEqual(entity.points.value, '2');
      assert.strictEqual(entity.points, 2);

      [entity] = await datastore.get(postKey);
      // without wrap numbers, expect to get a number '2'
      assert.strictEqual(entity.points, 2);

      // If we re-save here, without numbers wrapped, it will raise a warning
      // we have an unsafe to convert representation
      // NOTE: the data has been reuploaded in a different state than what was
      // retrieved
      await datastore.save(entity);
      [entity] = await datastore.get(postKey, { wrapNumbers: true });
      // expect as we saved, that this has been made into a a datastore int now.
      assert.strictEqual(entity.points.type, "DatastoreInt");
      assert.strictEqual(entity.points.value, "2");

    });
```

crwilcox:
The way I see it we have a few options how to proceed
1. do nothing, most users aren't going to have int-able doubles, those that do can always take care to wrap their numbers. There are issues here as get likes to auto-decode doubles for the user

2. add a warning. for this to work, we likely need to always wrap numbers coming back

3. discontinue the entire conversion layer. This degrades the experience for customers who simply use ints, or use non-int-able doubles.

For me option 3 is out due to the effects on all users, including those that have no risk of seeing this issue. that said, we have consistently for datastore and firestore apis seen that the layers to 'magically encode/decode' are a source of struggle. I think option 2 is good, but having numbers always come back wrapped is a sizable change as well. those types implement Number though so that likely would have a minimal impact on current users.

@vicb as someone who used this, is option 1 even an option? Was being informed that you can wrap your numbers specifically to certain types good enough? I figured not, but since we are at this point I figure it is always worth reasking :)

vicb:
Hey @crwilcox,

Below is my personal opinion.

I am trying to be constructive and this message has no intent to be (even passive) aggressive.

Referring to [your second to last message](https://github.com/googleapis/nodejs-datastore/pull/773#issuecomment-742847331):

> I think most of your comments are handled well enough

"well enough" sounds to me like a polite way to say they are not but that you won't do more.

>  I have no intent to write a utility to clean up mixed states across entities. I expect that most customers haven't even encountered this issue as it requires having 'int-able' doubles.

I think you are right to say "I expect" as this sounds like a wish.

My take on this is that all double are "int-able". It is only a matter of time before you get an int.
Which kind of make the issue even worse.

> Saving a gotten entity.

Your take on this is:

> not necessarily desired behavior.
> NOTE: this is a bit awkward as is since you can't prevent decoding of doubles

I think those comments capture my thought well. Nothing to add here.

I think you should make clear that your 2 lines are what you would expect as a "desired behavior" but not the actual behavior.
```
      // assert.strictEqual(entity.points.type, 'DatastoreDouble');
      // assert.strictEqual(entity.points.value, '2');
```

Referring to your [last message](https://github.com/googleapis/nodejs-datastore/pull/773#issuecomment-742849986)

> The way I see it we have a few options how to proceed

I assume you are referring top the the whole issue with saving/querying numbers.

> 1. do nothing

The problem is that customers pay for this DB.
I pay a few dozens $ per months to GCP for a hobby project.
There are probably company spending much more for a project critical to their business.
If it is the case it will be hard to tell them: we knew but we didn't fell like fixing this.

> 2. add a warning

It might be a solution.
But as of today I don't actually know who is able to understand this warning as there is no documentation.
IMO Documentation should be adder either before or at the same time as the warning.

Also all this wrapping (in/out) will be a pain for customers.

> 3. discontinue the entire conversion layer

Yes I think the layer in its current form should be discontinued as it is error prone.

Further thoughts.

I have proposed 2 things that I think are still valid:
1) [Propose/Discuss a few different solutions](https://github.com/googleapis/nodejs-datastore/pull/773#issuecomment-741420498). I think it would be valuable to wrote down the few options there are with their pros and cons.
2) [Adding a schema](https://github.com/googleapis/nodejs-datastore/issues/754#issuecomment-729864250) would allow automatic wrapping/unwrapping of the declared properties.

Lastly it would be nice to get an estimate of the complexity and cost of fixing this on the DB server side.

crwilcox:
> "well enough" sounds to me like a polite way to say they are not but that you won't do more.

While close, let me clarify what I meant. I think the warning covers this possibly destructive behavior to the point that that further effort would be of lower benefit than the required effort. I can expand.

1. My belief is that the utility to bring a db into a consistent column state is a largish work-item, adoption would likely be low, and given the nature of datastore, oddly prescriptive to our users. If you do believe that is in untrue and have the bandwidth, you are welcome to author it and PR here, or create a solution repo. We could add this to our documentation and perhaps users would find benefit in it.

2. The warning directs the user to two methods that will enforce a specific datastore type. Given this is a corner case issue, I also am not sure adding more documentation is worthwhile. If you have a thought where to add this documentation though I would be happy to incorporate it. We have a precedent for using warnings (see the large int warning in the same area) to inform customers of the possibly risky behavior. This pattern has worked so far.



Regarding this PR and addressing the issue of autoconversion causing friction, it seems to move this forward unwrapping of datastore types to numbers need to be have a reversed default. Currently we support specifying of wrapNumbers as part of ints, but by default we unbox them into Numbers. I think leaving support to do that is fine, but we likely want to opt to not unbox. `Datastore.int` and `Datastore.double` have the `Number` interface anyway, so there should be minimal disruption.

crwilcox:
https://github.com/googleapis/nodejs-datastore/pull/773/commits/1d80aeac4611c88e869a2c9a82f09e81ae9bb521 is further breaking, users who weren't specifying that we wrapNumbers, will now have it happen as the default was previously false. This change will break a great deal of tests, which I have not addressed yet.

This would fix #774 and allow us to improve the behavior of the above system test so that double/int values weren't so easily converted behind-the-scenes.

crwilcox:
Here is a modified version of that system test (committing after) with the behavior if we reverse wrapNumbers defaults

```
    it('should save/get an int-able double value via Datastore.', async () => {
      const postKey = datastore.key('Team');
      const points = Datastore.double(2);
      await datastore.save({ key: postKey, data: { points } });
      let [entity] = await datastore.get(postKey, { wrapNumbers: true });
      // Expect content is stored in datastore as a double with wrapping to 
      // return a wrapped double
      assert.strictEqual(entity.points.type, 'DatastoreDouble');
      assert.strictEqual(entity.points.value, 2);

      [entity] = await datastore.get(postKey, { wrapNumbers: false });
      // Verify that when requested with wrapNumbers false, we get a plain
      // javascript Number 2.
      assert.strictEqual(entity.points, 2);

      [entity] = await datastore.get(postKey);
      // Expect without any options, a wrapped double to be returned.
      assert.strictEqual(entity.points.type, 'DatastoreDouble');
      assert.strictEqual(entity.points.value, 2);

      // Save the data again, reget, ensuring that along the way it isn't 
      // somehow changed to another numeric type.
      await datastore.save(entity);
      [entity] = await datastore.get(postKey);
      // expect as we saved, that this property is still a DatastoreDouble.
      assert.strictEqual(entity.points.type, 'DatastoreDouble');
      assert.strictEqual(entity.points.value, 2);

    });
```

vicb:
> Regarding this PR and addressing the issue of autoconversion causing friction, it seems to move this forward unwrapping of datastore types to numbers need to be have a reversed default. Currently we support specifying of wrapNumbers as part of ints, but by default we unbox them into Numbers. I think leaving support to do that is fine, but we likely want to opt to not unbox. Datastore.int and Datastore.double have the Number interface anyway, so there should be minimal disruption.

Again I think that creating a document to discuss about the different solutions with their pros and cons, the impact for current users, and the impact for Google would help a lot here. And also I'd like to see in the document the option to fix this server side along with other solutions that all seem to require more changes by the customers.

stephenplusplus:
To match `Int`, should the custom Double class extend Number as well? Or in any case, a quick helper for some situations would be to add a `valueOf()` method to the custom Double class prototype, so that this would work:

```diff
assert.strictEqual(entity.points.type, 'DatastoreDouble');
assert.strictEqual(entity.points.value, 2);
+ assert.strictEqual(entity.points + 1, 3);
```

crwilcox:
@stephenplusplus 100%. There is some cleanup baggage if we go with this, though I didn't want to sink time into it until we are mostly sure the 'feel' of this would be right. I do think the source of our problems are in the automatic decoding of things as that then results in loss of type metadata for later operations.

@vicb Alterations to the ordering of objects in datastore is out of scope of the library. I also don't currently know of a plan to interleave double and ints. It may be worth opening an issue specifically about that in relation to the Datastore service.

Given that, I am focusing on the client library. In the near term, changes to the client library are likely more tractable to improving the end-user experience. Given that the number interface can be maintained, but type metadata as well if we don't auto unwrap, we can make this work much better I think :)

crwilcox:
@stephenplusplus I extended Double. Tests will very much need fixing if we go this route, though the user interface should be there.

crwilcox:
Tests have been updated to match behavior.

src/entity.ts
```diff
@@ -52,27 +52,44 @@ export namespace entity {
...
    * const datastore = new Datastore();
    * const aDouble = datastore.double(7.3);
    */
-  export class Double {
+  export class Double extends Number {
+    private _entityPropertyName: string | undefined;
     type: string;
     value: number;
-    constructor(value: number) {
+    constructor(value: number | string | ValueProto) {
```
vicb, line 66: Are Double / Int included in this type ?

src/entity.ts
```diff
@@ -52,27 +52,44 @@ export namespace entity {
...
+      super(typeof value === 'object' ? value.doubleValue : value);
```
vicb, line 67: What is the use case of constructing from a proto ?
If it needs to be supported int would probably make sense too.

src/entity.ts
```diff
@@ -52,27 +52,44 @@ export namespace entity {
...
+
       /**
        * @name Double#type
        * @type {string}
        */
       this.type = 'DatastoreDouble';
+
+      this._entityPropertyName =
```
vicb, line 75: Do we really need to keep the property name ?
If yes it would be good to comment why.

src/entity.ts
```diff
@@ -52,27 +52,44 @@ export namespace entity {
...
+        typeof value === 'object' ? value.propertyName : undefined;
+
       /**
        * @name Double#value
        * @type {number}
        */
-      this.value = value;
+      this.value =
```
vicb, line 82: Why do need a value member ?
It seems like it could be dropped as well as `valueOf` which are both already in the super class.

src/entity.ts
```diff
@@ -52,27 +52,44 @@ export namespace entity {
...
+        typeof value === 'object' ? Number(value.doubleValue) : Number(value);
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    valueOf(): any {
+      return Number(this.value);
+    }
+
+    toJSON(): any {
+      return {type: this.type, value: this.value};
```
vicb, line 92: This seems quite dangerous if the goal of `Double` is to fake a `Number` ?

src/entity.ts
crwilcox: Int has the same support from proto, and it is part of the entityFromEntityProto path. This code mirrors the behavior of our Int class.

src/entity.ts
crwilcox: Honestly, I am not sure what this is for, but I mirrored behavior in the Int implementation. Maybe @stephenplusplus recalls?

src/entity.ts
crwilcox: They ought to be, via the Number type. Though that isn't really a use path. Support for that could be added, but as you might detect a trend, this is a mirroring of the existing style we went with for Ints.

src/entity.ts
crwilcox: can you expand? Sorry, I don't quite understand/follow. Also, `class Int` has a similar toJson.

src/entity.ts
crwilcox: Possibly? This started as a string, but realized I could store this as a number. We store it as a string for Int mostly due to Int range differences.

Datastore Doubles have 64-bit double precision, IEEE 754, as do Node.js Numbers. So I was able to change this type and simplify a bit.

src/entity.ts
vicb: I am really not sure to get the intent of the change right but my understanding of the current state is that you want to return `Int`s and `Double`s that users would use as if they were `Number`s.

If I get this right then `Int`s and `Double`s should behave like `Number`s as much a possible.

However with this code:
- `JSON.stringify(1.23) == "1.23"`
- but `JSON.stringify(Double(1.23)) == '{"type": ..., "value": ...}'`

While not related to `toJSON()`  returning wrapped numbers (`Int` / `Double`) could be problematic:

```
const [entity] = await ds.get(...);

// entity.price is a Double, i.e. Double(10)

entity.price *= 0.90;

// entity.price is now a number (9, "int")

// warning + DB screwed
await ds.save(entity);
```

vicb (commented):
I added some details on my former brief review.

But at this point I am not sure any more sure of how exactly you are trying to solve the number thing. It might not be applicable ?

src/entity.ts
```diff
@@ -52,27 +52,44 @@ export namespace entity {
...
        * @type {number}
        */
-      this.value = value;
+      this.value =
+        typeof value === 'object' ? Number(value.doubleValue) : Number(value);
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    valueOf(): any {
+      return Number(this.value);
```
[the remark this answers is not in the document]
bcoe: Playing around with extending `Number`, there isn't access to a `this.value` unless we store the value ourselves:

```js
class F extends Number {
  constructor(value) {
    super(value)
  }
  blerg() {
    console.info(this.value)
  }
}

const f = new F(99);
console.info(f.valueOf())
f.blerg()
```

outputs:

```
99
undefined
```

~Part of me wonders if we should avoid extending the base `Number` type, and instead create a `DatastoreNumber` type, since `Number` implies a a larger interface than we potentially want to support out of the gate?~

Edit: I see we're already extending `Number` in `Int`, I'd stick with this approach.

src/entity.ts
bcoe: Since we're debouncing the typescast warning any ways, I'd consider just dropping the if and printing both.

test/entity.ts
```diff
@@ -19,6 +19,7 @@ import * as sinon from 'sinon';
 import {Datastore} from '../src';
 import {Entity} from '../src/entity';
 import {IntegerTypeCastOptions} from '../src/query';
+import {decode} from 'punycode';
```
bcoe, line 22: Is this used anywhere, I think it might just be `vscode` adding something?

src/entity.ts
bcoe: feels kind of like this should be private.

system-test/datastore.ts
```diff
@@ -268,6 +269,40 @@ describe('Datastore', () => {
       await datastore.delete(postKey);
     });
 
+    it('should save/get an int-able double value via Datastore.', async () => {
+      const postKey = datastore.key('Team');
+      const points = Datastore.double(2);
+      await datastore.save({key: postKey, data: {points}});
+      let [entity] = await datastore.get(postKey, {wrapNumbers: true});
```
bcoe, line 276: As a TODO, I wish we had unit tests for the Int and Double types, vs., only integration tests.

bcoe (changes requested):
Mainly left a few nits, I think modeling after `class Int` is the right approach.

src/entity.ts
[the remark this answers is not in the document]
vicb:
```typescript
class F extends Number {
  constructor(value) {
    super(value);
  }
  blerg() {
    console.info(Number(this));
  }
}

const f = new F(99);
f.blerg(); // outputs 99 as expected
```

src/entity.ts
vicb: I haven't checked the use cases but it seems dangerous

if an entity has the form `{oldValue: Int, newValue: Int}` and you do (with wrapped Number):

```typescript
const [entity] = await ds.get(...);

entity.oldValue = entity.newValue;
entity.newValue = ds.Int(123);

await ds.save(entity);
```

I can only imagine that keeping the entity property name would cause problems as `oldValue` would have a reference to `"newValue"`.

I am not sure if it is actually a problem in the current code but at least it makes it easy to shoot yourself in the foot.

meredithslota:
Coming back to this — I see in https://github.com/googleapis/nodejs-datastore/pull/773#issuecomment-742913844 a request to create a document describing potential solutions (including server-side solutions) and their pros/cons. Did such a document ever get authored? It feels like perhaps we aren't yet aligned on the path forward yet, but I can't tell if there was a decision reached outside of this thread.

crwilcox:
There isn't much in the way of server side to be done as datastore sorts by name and type, and javascript doesn't distinguish between the different numerical types. The path forward almost certainly will break things for users. I didn't reach out to the datastore backend team about this as I don't believe there is an issue with the way the backend operates.

test/entity.ts
danieljbruce: Not used. Will delete.

system-test/datastore.ts
danieljbruce: I think a good unit test corresponding to this one would be to mock out the `get` and the `save` and ensure that getting a Double/Int leads to saving the same type.

danieljbruce (commented):
Adding comments now I plan to follow up on.

src/entity.ts
danieljbruce: I think printing both could get noisy as Chris said. I ended up going with @feywind's suggestion here.

danieljbruce (commented):
Left a bunch of comments for more suggestions.

danieljbruce:
A continuation of the work done on this PR is contained in https://github.com/googleapis/nodejs-datastore/pull/941/files. Closing for now, but feel free to leave any comments for on the new PR for next steps.

PR #788, opened 2021-02-01
fix: check property existence for excludeFromIndexes with wildcard

### LINKED ISSUE #787: excludeFromIndexes using wildcard  on non-existing data causes `TypeError: Cannot read property 'entityValue' of undefined`.
`excludeFromIndex` using wildcard `.*` on non-existing data causes `TypeError: Cannot read property 'entityValue' of undefined`.

For example,

```typescript
datastore.save(
  key,
  {
    prop: 'dummy',
  },
  excludeFromIndexes: [
    // this just ignored
    'non_exist_property',
    // causes TypeError: Cannot read property 'entityValue' of undefined`
    // should also be ignored
    'non_exist_property.*' 
  ]
)
```

As `non_exist_property` in `excludeFromIndexes` is ignored, `non_exist_property.*` should be ignored and we should be able to add data.

---

This causes a problem when we try to exclude some **optional** properties like

```typescript
interface Example {
  prop1: string,
  prop2?: {
    nested1: string,
    nested2: string
  }
}
```

and want to exclude `prop2.*` from indexes.

#### Environment details

  - OS: OSX (but it should happen on any OS)
  - Node.js version: 10.22.1 (but it should happen on any versions)
  - npm version: 6.14.6 (dit)
  - `@google-cloud/datastore` version: `6.3.1`

#### Steps to reproduce

```typescript
datastore.save(
  key,
  {
    prop: 'dummy',
  },
  excludeFromIndex: [
    // this just ignored
    'non_exist_property',
    // causes TypeError: Cannot read property 'entityValue' of undefined`
    // should also be ignored
    'non_exist_property.*' 
  ]
)
```

##### failing test case
```diff
diff --git a/test/index.ts b/test/index.ts
index 42bb76d..f31b52a 100644
--- a/test/index.ts
+++ b/test/index.ts
@@ -1677,6 +1677,8 @@ describe('Datastore', () => {
             'metadata.otherProperty',
             'metadata.obj.*',
             'metadata.longStringArray[].*',
+            'undefinedData.*',
+            'undefinedArray[].*',
           ],
         },
         assert.ifError
```

Related https://github.com/googleapis/nodejs-datastore/pull/451

tanishiking:
Fixes https://github.com/googleapis/nodejs-datastore/issues/787 🦕

As described in https://github.com/googleapis/nodejs-datastore/issues/787, `nodejs-datastore` raises `TypeError: Cannot read property 'entityValue' of undefined` if we try to exclude **undefined** properties from index using `.*`.

This is because we didn't check the property existence (e.g. for `non_exist_data.*`, we should check `non_exist_data`'s existence), so this PR make `entity.ts` check their existence before traversing data for `excludeFromIndexes` with a wildcard.

---


tanishiking:
oops CLA signed

tanishiking:
@googlebot I signed it!

JustinBeckwith:
@crwilcox or @stephenplusplus could I trouble y'all to have a look here?

udnes99:
Can someone please add this fix already? It is quite the detrimental bug.

zomgbre:
It is a huge shame that this is still broken today in 7.0.0. In fact, two years ago my colleague posted about the same issue and the issue was closed with inaccurate information.

src/entity.ts
```diff
@@ -882,15 +883,17 @@ export namespace entity {
...
-        const array = entity.properties![firstPathPart].arrayValue;
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        array.values.forEach((value: any) => {
-          if (value.entityValue) {
-            excludePathFromEntity(value.entityValue, '.*');
-          } else {
-            excludePathFromEntity(value, '');
-          }
-        });
+        if (entity.properties![firstPathPart]) {
```
danieljbruce, line 886: So we just check to see that this property is defined before we move further as that prevents the error from being thrown. Makes sense. There seems to be no value in further traversing a path that the entity will not have any corresponding properties for.

src/entity.ts
```diff
@@ -857,6 +857,7 @@ export namespace entity {
       if (
         firstPathPartIsArray &&
         // check also if the property in question is actually an array value.
+        entity.properties![firstPathPart] &&
```
danieljbruce, line 860: I see no harm in this. If we don't check for this and `entity.properties![firstPathPart]` ever happens to be undefined then this will throw an error and I can't imagine that being desirable.

test/index.ts
```diff
@@ -1677,6 +1677,8 @@ describe('Datastore', () => {
             'metadata.otherProperty',
             'metadata.obj.*',
             'metadata.longStringArray[].*',
+            'undefinedData.*',
+            'undefinedArray[].*',
```
danieljbruce, line 1681: We shouldn't modify the existing test. We should make a new test that tests for these specific values so we don't lose any existing test coverage.

danieljbruce (commented):
Looks fine, but a small difference is required in the way we test this feature.

danieljbruce:
Closed in favour of https://github.com/googleapis/nodejs-datastore/pull/1114.

PR #848, opened 2021-06-28
fix: Resolve getProjectId

### LINKED ISSUE #844: getProjectId never resolves
#### Environment details

  - OS: Ubuntu 20.04.2 LTS 
  - Node.js version: v14.17.0
  - npm version: 6.14.13
  - `@google-cloud/datastore` version: 6.4.1

#### Steps to reproduce

Run the following code:
```js
const { Datastore } = require('@google-cloud/datastore');

const datastore = new Datastore();
datastore.getProjectId()
  .then(projectId => console.log(`projectId: ${projectId}`))
  .catch(err => console.log(`err ${err}`));
```

Expected output is `projectId: <my-project-id>` or `err <error>` but no output is shown.

### Workaround

I had a quick look and it seems like `getProjectId` is promisified even though it already returns a promise. I added `getProjectId` to the promisify exclude list in `/node_modules/@google-cloud/datastore/build/src/index.js:1484`:
```js
promisify_1.promisifyAll(Datastore, {
    exclude: [
        'double',
        'isDouble',
        'geoPoint',
        'isGeoPoint',
        'index',
        'int',
        'isInt',
        'createQuery',
        'key',
        'isKey',
        'keyFromLegacyUrlsafe',
        'transaction',
        'getProjectId'
    ],
});
```
Also found at: https://github.com/googleapis/nodejs-datastore/blob/ccc466a5fce8cb78a0980f961bfa473c56a12c38/src/index.ts#L1747-L1762

With `getProjectId` excluded, the output is `projectId: <my-project-id>` which is the expected output.

billyjacobson:
- [ ] Ensure the tests and linter pass

Fixes #844 🦕

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1749,6 +1749,7 @@ promisifyAll(Datastore, {
     'double',
     'isDouble',
     'geoPoint',
+    'getProjectId',
     'isGeoPoint',
     'index',
     'int',
--- a/system-test/datastore.ts
+++ b/system-test/datastore.ts
@@ -60,6 +60,11 @@ describe('Datastore', () => {
     assert.ok(keys);
   });
 
+  it('should get the project id', async () => {
+    const projectId = await datastore.getProjectId();
+    assert.notEqual(projectId, null);
+  });
+
   describe('create, retrieve and delete', () => {
     const post = {
       title: 'How to make the perfect pizza in your grill',


billyjacobson:
Hey I'm not sure how to test this but was just doing a solution based on the GitHub issue

src/index.ts
```diff
@@ -1749,6 +1749,7 @@ promisifyAll(Datastore, {
     'double',
     'isDouble',
     'geoPoint',
+    'getProjectId',
```
JustinBeckwith, line 1752: It looks like this makes an actual change to the source of the library - can you use the `fix:` conventional commit prefix to make sure this triggers a launch?

system-test/datastore.ts
```diff
@@ -60,6 +60,11 @@ describe('Datastore', () => {
     assert.ok(keys);
   });
 
+  it('should get the project id', async () => {
+    const projectId = await datastore.getProjectId();
+    assert.notEqual(projectId, null);
```
JustinBeckwith, line 65: It could still come back as `undefined` here, or an empty string :/  What do you think of `assert.ok(projectId)` to capture falsey values?

JustinBeckwith (approved):
LGTM w/nits

system-test/datastore.ts
billyjacobson: I'm fine with that. I'd just assume that for the other tests to pass the projectid would need to be defined

crwilcox:
The cause of the test failure was that it seems import/export uses a testing project that didn't have legacy reader access to the bucket we use. This is likely related to an issue that happened ~2 weeks back where the test bucket was erroneously deleted. When recreated, perms for this extra account, `billy-testing`, were not granted.

PR #882, opened 2021-10-11
chore(cloud-rad): fix links in docs

fhinkel:
See internally b/179484422

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -155,10 +155,7 @@ const urlSafeKey = new entity.URLSafeKey();
  * you don't need to provide a `projectId`.
  *
  *
- * @class
- * @extends {DatastoreRequest}
- *
- * @see [Cloud Datastore Concepts Overview]{@link https://cloud.google.com/datastore/docs/concepts/overview}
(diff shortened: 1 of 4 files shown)


JustinBeckwith (approved):
Do we have an automated way to detect these kinds of failures?  I'm guessing there are more cases of this about, and it would be a nice thing to have the link checker verify.

fhinkel:
> Do we have an automated way to detect these kinds of failures? I'm guessing there are more cases of this about, and it would be a nice thing to have the link checker verify.

Is this something we'd implement in `gts`?

JustinBeckwith:
Today with the `jsdoc` based approach, we run the `docs-test` as part of every PR.  This test:
- Uses `jsdoc` to generate the exact HTML we're going to host on googleapis.dev
- Runs `linkinator` against the generated docset to look for dead links

If there's an easy way to get something _mostly_ like Cloud RAD generating HTML output as part of the CI job, that would be ideal because we could do the exact same kind of link scanning.

Separately, if we're looking for tsdoc style linting - an `eslint` plugin feels appropriate.  Not sure if such a plugin exists, or if that's something we'd need to write.

PR #936, opened 2022-04-19
feat: Query filters for datastore

danieljbruce:
This PR supports three more types of query filters and provides tests for those new filters too to ensure they can be used in datastore.


test/query.ts
```diff
@@ -75,7 +75,10 @@ describe('Query', () => {
         .filter('name', '=', 'Title')
         .filter('count', '>', 20)
         .filter('size', '<', 10)
-        .filter('something', '>=', 11);
+        .filter('something', '>=', 11)
```
kolea2, line 78: lgtm, please at some system tests as well

kolea2 (changes requested):

kolea2 (approved):

PR #956, opened 2022-06-22
feat: Added count filter

danieljbruce:
With this pull request we can use the client to do a query filter based on the COUNT filter which we intend to support.


danieljbruce:
Having done this, I don't think this is right... I think we want to leverage the aggregateValue. More research is needed here...

danieljbruce:
This PR is no longer relevant. COUNT is an aggregate method.

PR #963, opened 2022-07-18
feat: Snapshot reads

danieljbruce:
This implements the snapshot read feature which allows the user to get data from a previous point in time.

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -279,9 +279,11 @@ class DatastoreRequest {
 
       if (options.consistency) {
         const code = CONSISTENCY_PROTO_CODE[options.consistency.toLowerCase()];
+        const readTime = options.readTime;
 
         reqOpts.readOptions = {
           readConsistency: code,
+          readTime,
         };
       }
 
@@ -1058,7 +1060,7 @@ export interface RequestConfig {
 export interface RequestOptions {
   mutations?: google.datastore.v1.IMutation[];
   keys?: Entity;
-  readOptions?: {readConsistency?: number; transaction?: string};
+  readOptions?: {readConsistency?: number; transaction?: string, readTime?: number};
   partitionId?: google.datastore.v1.IPartitionId | null;
   transactionOptions?: {
     readOnly?: {};
--- a/src/request.ts
+++ b/src/request.ts
@@ -54,6 +54,7 @@ import {
   RunQueryCallback,
 } from './query';
 import {Datastore} from '.';
+import ITimestamp = google.protobuf.ITimestamp;
 
 /**
  * A map of read consistency values to proto codes.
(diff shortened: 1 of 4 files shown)


src/request.ts
```diff
@@ -284,6 +285,16 @@ class DatastoreRequest {
           readConsistency: code,
         };
       }
+      if (options.readTime) {
+        if (!reqOpts.readOptions) {
```
bcoe, line 289: nit:

I might make this explicitly, `if (reqOpts.readOptions !== undefined)`.

Since JavaScript is so loose with its truthiness, I like to be explicit.

src/request.ts
```diff
@@ -284,6 +285,16 @@ class DatastoreRequest {
...
+          reqOpts.readOptions = {};
+        }
+        const readTime = options.readTime;
+        const seconds = readTime / 1000;
```
bcoe, line 293: Could you just make this `parseInt(readTime / 1000`, rather than having the floor? (I'm guessing this type should be an integer?

bcoe (approved):
Left a couple small nits.

src/request.ts
danieljbruce: Changed to `if (reqOpts.readOptions === undefined) {`

src/request.ts
danieljbruce: `parseInt` only accepts string as a parameter. Are you sure about this? Should I convert it into a string and then use parseInt?

danieljbruce (commented):
First follow-up

system-test/datastore.ts
```diff
@@ -308,6 +308,39 @@ describe('Datastore', () => {
       await datastore.delete(postKey);
     });
 
+    // TODO: Modify this
```
kolea2, line 311: still needed?

system-test/datastore.ts
```diff
@@ -308,6 +308,39 @@ describe('Datastore', () => {
...
+        isDraft: false,
+        wordCount: 400,
+        rating: 5.0,
+        likes: null,
+        metadata: {
+          views: 100,
+        },
+      };
+      const path = ['Post', 'post1'];
+      const postKey = datastore.key(path);
```
kolea2, line 330: nit - I would rename this to `postKey1` or something similar to reference this is related to post1 rather than post2

system-test/datastore.ts
```diff
@@ -308,6 +308,39 @@ describe('Datastore', () => {
...
+      await datastore.save({key: postKey, data: post});
+      await sleep(1000);
+      const savedTime = Date.now();
+      await sleep(1000);
+      // Save new post2 data, but then verify the timestamp read has post1 data
+      await datastore.save({key: postKey, data: post2});
+      const [entity] = await datastore.get(postKey, {readTime: savedTime}); // Include new options here
```
kolea2, line 337: maybe it would be nice to do two of these; one for each post

system-test/datastore.ts
danieljbruce: Should be removed. Good catch!

system-test/datastore.ts
danieljbruce: The same `postKey` is used for both `post1` and `post2`. There is only one `postKey`.

system-test/datastore.ts
danieljbruce: Good plan.

danieljbruce: Done

PR #969, opened 2022-08-23
Count aggregation 2

danieljbruce:
This PR implements the new COUNT feature when querying datastore which returns a count of the number of records instead of the records themselves.


kolea2:
Closing in favor of https://github.com/googleapis/nodejs-datastore/pull/972

PR #972, opened 2022-08-26
feat: Add the count aggregate function

danieljbruce:
This PR is just like the other PR on count aggregation, but it contains only the relevant changes to make the count filter work.


danieljbruce:
This is a copy of https://github.com/googleapis/nodejs-datastore/pull/969, but it only contains the changes needed once more code is merged with main.

src/request.ts
```diff
@@ -744,27 +745,94 @@ class DatastoreRequest {
...
+            {
+              client: 'DatastoreClient',
+              method: 'runQuery',
+              reqOpts,
+              gaxOpts: options.gaxOptions,
+            },
+            onResultSet
+        );
+      } else {
+        const aggregationQueryOptions: AggregationQueryOptions = {
```
kolea2, line 772: why not separate this out into a different method? `runAggregationQueryStream` vs `runQueryStream`?

danieljbruce: I see what you mean. This function is getting gnarly.

So currently this code is being run in `runQueryStream`, if we keep `runQueryStream` the way it was before and add a new function `runAggregationQueryStream` then there are a few downsides:
1. We expand the API surface by adding a new function
2. With a new function the user has to know to pass an aggregate query into `runAggregationQueryStream` and a regular query into `runQueryStream` if they are using requests directly. Note that both aggregate and non-aggregate queries are both instances of the `Query` class so there could be situations where it is not clear which type of query we are dealing with.

Although, maybe what you were getting at here was to move the code in the `else` branch out into a separate, private function (ie. `runAggregationQueryStream`) in a utility function which I think is a good idea. It might be a good idea with the `if` code as well.

danieljbruce: I think what we could do here is add two inner functions, one for the `if` and one for the `else` so that the if/else block is concise. Do you think that would address the concern in question?

danieljbruce: See comment above

src/request.ts
```diff
@@ -744,28 +746,112 @@ class DatastoreRequest {
...
-          reqOpts,
-          gaxOpts: options.gaxOptions,
-        },
-        onResultSet
-      );
+      if (query.aggregations.length === 0) {
+        runDataQuery(sharedQueryOpts, queryProto);
+      } else {
+        runAggregationQuery(sharedQueryOpts, queryProto);
+      }
```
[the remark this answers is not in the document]
danieljbruce: This comment is out of date.

src/aggregate.ts
```diff
@@ -0,0 +1,77 @@
...
+  type = AGGREGATE_QUERY;
+  aggregations: Array<AggregateField>;
+  query: Query | undefined;
+
+  constructor(query: Query) {
+    this.query = query;
+    this.aggregations = [];
+  }
+
+  count(alias: string, upTo: number): AggregateQuery {
```
kolea2, line 32: note - need to remove upTo

src/aggregate.ts
```diff
@@ -0,0 +1,77 @@
...
+  alias(alias: string): AggregateField {
+    this.alias_ = alias;
+    return this;
+  }
+
+  abstract toProto(): any;
+}
+
+class Count extends AggregateField {
+  upTo_?: number;
```
kolea2, line 63: same here

kolea2 (changes requested):

src/aggregate.ts
```diff
@@ -0,0 +1,69 @@
...
+    this.aggregations = [];
+  }
+
+  count(alias: string): AggregateQuery {
+    this.aggregations.push(AggregateField.count().alias(alias));
+    return this;
+  }
+
+  aggregate(aggregates: AggregateField[]): AggregateQuery {
+    aggregates.forEach(aggregate => this.aggregations.push(aggregate));
```
ruyadorno, line 38:
```suggestion
    for (const aggregate of aggregates) {
      this.aggregations.push(aggregate);
    }
```

[optional] this change would avoid having a unnecessary returning statement from the arrow function being passed to `forEach` while also avoiding a long one-liner expression

src/aggregate.ts
```diff
@@ -0,0 +1,69 @@
...
+    return this;
+  }
+
+  toProto(): any {
```
ruyadorno, line 42: maybe add some comments to disable the linting for all `toProto` methods?

src/aggregate.ts
```diff
@@ -0,0 +1,69 @@
...
+    return this.aggregations.map(aggregation => aggregation.toProto());
+  }
+}
+
+abstract class AggregateField {
+  alias_?: string;
```
ruyadorno, line 48: [optional] `alias_` could also be defined as a Symbol, given a `const _alias = Symbol('alias');` definition, e.g:

```suggestion
  [_alias]?: string;
```

src/aggregate.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+class Count extends AggregateField {
+  toProto(): any {
+    const count = Object.assign({});
+    return Object.assign({count}, this.alias_ ? {alias: this.alias_} : null);
+  }
+}
+
+export {AggregateField, AggregateQuery, AGGREGATE_QUERY};
```
ruyadorno, line 69: [optional] I'm not sure about the architectural/design choices here but I have the impression that the it would be simpler for this file to only expose a `count` function and avoid exposing this abstract `AggregateField` class.

src/index.ts
```diff
@@ -64,6 +64,7 @@ import {
 import {Transaction} from './transaction';
 import {promisifyAll} from '@google-cloud/promisify';
 import {google} from '../protos/protos';
+import {AggregateField, AggregateQuery} from './aggregate';
```
ruyadorno, line 67: address linting issue

src/query.ts
```diff
@@ -21,6 +21,7 @@ import {Entity} from './entity';
 import {Transaction} from './transaction';
 import {CallOptions} from 'google-gax';
 import {RunQueryStreamOptions} from '../src/request';
+import {AggregateField, AggregateQuery} from './aggregate';
```
ruyadorno, line 24:
```suggestion
```

ruyadorno (commented):
Left a few comments, mostly optional improvements to the implementation.

src/aggregate.ts
telpirion: Acknowledged. There are design reasons for exposing this field

src/aggregate.ts
telpirion: Acknowledged.

src/index.ts
telpirion: Done.

telpirion (approved):

ferreiro:
hey, when are you planning on shipping this? I want to start using it. @telpirion @danieljbruce

kolea2 (changes requested):
need to make changes in https://github.com/googleapis/nodejs-datastore/blob/main/src/transaction.ts

src/request.ts
```diff
@@ -564,6 +566,78 @@ class DatastoreRequest {
...
+    options: RunQueryOptions,
+    callback: RequestCallback
+  ): void;
+  runAggregateQuery(query: AggregateQuery, callback: RequestCallback): void;
+  runAggregateQuery(
+    query: AggregateQuery,
+    optionsOrCallback?: RunQueryOptions | RequestCallback,
+    cb?: RequestCallback
+  ): void | Promise<RunQueryResponse> {
+    // TODO: Check that the query was provided in `over`.
```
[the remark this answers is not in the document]
danieljbruce: These TODOs are not relevant anymore and have been removed.

src/aggregate.ts
```diff
@@ -0,0 +1,94 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// ** This file is automatically generated by gapic-generator-typescript. **
+// ** https://github.com/googleapis/gapic-generator-typescript **
+// ** All changes to this file may be overwritten. **
```
Mariatta, line 17: Is the notice correct that this is autogenerated?

danieljbruce: It is not autogenerated. Thanks for catching this one! Changes pushed.

system-test/datastore.ts
```diff
@@ -797,7 +798,54 @@ describe('Datastore', () => {
       const [entities] = await datastore.runQuery(q);
       assert.strictEqual(entities!.length, 6);
     });
-
+    describe('with a count filter', () => {
```
kolea2, line 801: Looking good! Please add some tests using `limit`

kolea2 (dismissed):

system-test/datastore.ts
danieljbruce: Added 2 tests using limit

ruyadorno (approved):

Mariatta (approved):

PR #1061, opened 2023-01-19
feat: introduce EntityFilter class with support for and/or filters

danieljbruce:
This code will enable the composite OR filters (much like the same way the composite AND filters work).


system-test/datastore.ts
```diff
@@ -798,6 +799,83 @@ describe('Datastore', () => {
...
+        const q = datastore
+          .createQuery('Character')
+          .filter(
+            Filter.OR([
+              new PropertyFilter('family', '=', 'Stark'),
+              new PropertyFilter('appearances', '>=', 20),
+            ])
+          );
+        const [entities] = await datastore.runQuery(q);
+        assert.strictEqual(entities!.length, 6);
```
danieljbruce, line 850: This will change.

src/filter.ts
```diff
@@ -0,0 +1,159 @@
...
+}
+
+/**
+ * A Filter is a class that contains data for a filter that can be translated
+ * into a proto when needed.
+ *
+ * @see {@link https://cloud.google.com/datastore/docs/concepts/queries#filters| Filters Reference}
+ *
+ */
+export abstract class Filter {
```
danieljbruce, line 38: Maybe this should be an interface.

system-test/datastore.ts
```diff
@@ -811,6 +812,83 @@ describe('Datastore', () => {
...
+      describe('using hasAncestor and Filter class', () => {
+        const secondAncestor = datastore.key([
+          'Book',
+          'GoT',
+          'Character',
+          'Rickard',
+          'Character',
+          'Eddard',
+        ]);
+        it('should run a query using hasAncestor first', async () => {
```
danieljbruce, line 874: typo first/last should be switched around

system-test/datastore.ts
```diff
@@ -811,6 +812,83 @@ describe('Datastore', () => {
...
+        const q = datastore
+          .createQuery('Character')
+          .filter(
+            AND([
+              new PropertyFilter('family', '=', 'Stark'),
+              new PropertyFilter('appearances', '>=', 20),
+            ])
+          );
+        const [entities] = await datastore.runQuery(q);
+        assert.strictEqual(entities!.length, 6);
```
Mariatta, line 851: Is it possible to also assert that the resulting entities indeed have the values `family = Stark` and `apprearances >= 20`?

system-test/datastore.ts
```diff
@@ -811,6 +812,83 @@ describe('Datastore', () => {
...
+        const q = datastore
+          .createQuery('Character')
+          .filter(
+            OR([
+              new PropertyFilter('family', '=', 'Stark'),
+              new PropertyFilter('appearances', '>=', 20),
+            ])
+          );
+        const [entities] = await datastore.runQuery(q);
+        assert.strictEqual(entities!.length, 8);
```
Mariatta, line 863: Is it possible to assert that in resulting query, there are entities that have value `family = Stark` and not `appearances >=20`?

system-test/datastore.ts
danieljbruce: I looked at the data and there is actually one entity that has family `['Stark', 'Tully']`. I believe this is intended behaviour for the filter because this is also the result when we test a case where legacy `.filter` is used.

danieljbruce: I just checked `should filter queries with defined indexes` on main. It seems this filter checks to see if `Stark` is present. I guess this matches what you said. We can check for its pressence.

system-test/datastore.ts
danieljbruce: Right now I am seeing that all the entities which are returned have a family name of Stark.

src/entity.ts
```diff
@@ -1249,34 +1238,12 @@ export namespace entity {
       queryProto.startCursor = query.startVal;
     }
 
-    if (query.filters.length > 0) {
-      const filters = query.filters.map(filter => {
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        let value: any = {};
-
-        if (filter.name === '__key__') {
```
kolea2, line 1257: why are we removing this `__key__` logic?

src/entity.ts
```diff
@@ -1249,34 +1238,12 @@ export namespace entity {
...
-          filters,
-          op: 'AND',
-        },
-      };
+    if (query.filters.length > 0 || query.newFilters.length > 0) {
+      const filters = query.filters.map(
+        filter => new PropertyFilter(filter.name, filter.op, filter.val)
+      );
+      const newFilters = query.newFilters;
+      queryProto.filter = AND(newFilters.concat(filters)).toProto();
```
kolea2, line 1246: a little confused by this, what is this logic doing? Can you add some comments?

src/query.ts
```diff
@@ -201,24 +207,29 @@ class Query {
    * const keyQuery = query.filter('__key__', key);
    * ```
    */
-  filter(property: string, value: {} | null): Query;
-  filter(property: string, operator: Operator, value: {} | null): Query;
+  filter(propertyOrFilter: string | NewFilter, value?: {} | null): Query;
```
kolea2, line 210: why can't this just be `string | Filter`?

system-test/datastore.ts
danieljbruce: Done in last commit

system-test/datastore.ts
danieljbruce: Done in last commit

src/entity.ts
danieljbruce: This OR query implementation created duplicate code so this code has not been removed, it has only been refactored into the `PropertyFilter` object.

Take a look at https://github.com/googleapis/nodejs-datastore/pull/1061/files#diff-87404c91a22dbabcc334f9a5288341281cb2b6625533e1207efcac16dc1e2ffdR77.

src/query.ts
```diff
@@ -76,6 +76,7 @@ class Query {
   namespace?: string | null;
   kinds: string[];
   filters: Filter[];
+  newFilters: NewFilter[];
```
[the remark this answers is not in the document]
danieljbruce: At first I put all the filters into the `filters` object whether they were the old type of filter or the new filter class. This caused a compiler error however caught by https://github.com/googleapis/nodejs-datastore/pull/1067/files#diff-a9fb5a76e455ff1f3d63199cae37a32e289867ec02b1ec205b79468d074a211eR881.

Currently, users may be popping off the `filters` variable and expecting that variable to have the `val` property, but it won't if we allow `filter` to store the new `Filter` class.

src/query.ts
danieljbruce: In this module, `NewFilter` is just what we call the imported `Filter` class. We cannot call it `Filter` because the name `Filter` is already taken in this module.

https://github.com/googleapis/nodejs-datastore/pull/1061/files#diff-ad0978c0145d3620d2a97bdff4307f9496529c0e3011a845e2d5bf1a169d2538R21

src/entity.ts
danieljbruce: Added comments to the code here.

This is mainly just a refactor. Old code is being moved into the `Filter` classes, but this new code does the same thing as the removed code. The functionality had to be pulled into the `Filter` classes to avoid creating duplicate code.

system-test/datastore.ts
danieljbruce: Solved

src/query.ts
danieljbruce: Filter has been renamed `entityFilter` so as not to cause confusion.

src/query.ts
```diff
@@ -170,7 +176,7 @@ class Query {
    *
    * @see {@link https://cloud.google.com/datastore/docs/concepts/queries#datastore-property-filter-nodejs| Datastore Filters}
    *
-   * @param {string} property The field name.
+   * @param {string | NewFilter} propertyOrFilter The field name.
```
kolea2, line 179: shouldn't this be EntityFilter now? (side note - how is this compiling?)

kolea2 (changes requested):

src/query.ts
danieljbruce: You are right. This needs to change. This is compiling because this is actually just a documentation comment.

danieljbruce: Done

src/entity.ts
```diff
@@ -21,6 +21,7 @@ import {PathType} from '.';
 import {protobuf as Protobuf} from 'google-gax';
 import * as path from 'path';
 import {google} from '../protos/protos';
+import {AND, PropertyFilter} from './filter';
```
bcoe, line 24: nit, upper case is usually reserved for variable constants `const FOO = 9;  BAR = 99`, to make it clear that this is an exported method, I might go with:

`and`/`or`,  or `applyAnd`, `applyOr`.

src/filter.ts
```diff
@@ -0,0 +1,159 @@
...
+    return value;
+  }
+
+  /**
+   * Gets the proto for the filter.
+   *
+   */
+  // eslint-disable-next-line
+  toProto(): any {
+    const OP_TO_OPERATOR = new Map([
```
bcoe, line 91: nit: I would probably hoist this to the top of the file, rather than declaring in `toProto`, at scale this will speed things up since you're allocating less `Maps`.

src/query.ts
```diff
@@ -201,24 +207,29 @@ class Query {
...
+    propertyOrFilter: string | EntityFilter,
+    operatorOrValue?: Operator,
     value?: {} | null
   ): Query {
-    let operator = operatorOrValue as Operator;
-    if (arguments.length === 2) {
-      value = operatorOrValue as {};
-      operator = '=';
-    }
+    if (isFilter(propertyOrFilter)) {
```
bcoe, line 217: This is good use of TypeScript's `is` functionality 👍

bcoe (approved):
I left a few minor suggestions, this is looking really good IMO:

- A comprehensive suite of tests.
- Code reads well.

bcoe:
@danieljbruce the tile of this PR seems a bit off, I believe you're actually introducing the ability to filter using a class rather than just a string?

Perhaps a better title:

```
feat: introduce EntityFilter class with support for and/or filters
```

Or something like this.

src/entity.ts
danieljbruce: Choosing the former

src/query.ts
danieljbruce: Thanks

src/entity.ts
```diff
@@ -1249,34 +1238,20 @@ export namespace entity {
...
-          op: 'AND',
-        },
-      };
+    // Check to see if there is at least one type of legacy filter or new filter.
+    if (query.filters.length > 0 || query.entityFilters.length > 0) {
+      // Convert all legacy filters into new property filter objects
+      const filters = query.filters.map(
+        filter => new PropertyFilter(filter.name, filter.op, filter.val)
+      );
+      const newFilters = query.entityFilters;
```
kolea2, line 1247: nit - rename variable

kolea2 (approved):

src/entity.ts
danieljbruce: This was done

Mariatta (approved):

PR #1065, opened 2023-02-14
fix: Query filter on __key__ using the "IN" operator

### LINKED ISSUE #1024: Query filter on __key__ using the "IN" operator
#### Context

The datastore API allows filtering on `__key__` with the "IN" operator and an array of Keys as value.  
I want to do using this library.

#### Problem description

This lib does not allow to query entities filtering on __key__ using the "IN" operator, and throws:  
`InvalidKey: A key should contain at least a kind.`  
=> The code expects a Key when I send an array of Keys.

#### Steps to reproduce

```js
const entitiesKey = [
  datastore.key({
    namespace,
    path: [kind, '01ece338-066d-443b-8c33-15414f971092']
  }),
  datastore.key({
    namespace,
    path: [kind, '0750f7ed-caf5-44d4-9105-0d72a81eeb9b']
  })
]
const query = datastore.createQuery(namespace, kind)
query.filter('__key__', 'IN', entitiesKey)
const entities = datastore.runQuery(query)
```

Throwing this:
```log
InvalidKey: A key should contain at least a kind.
    at Object.keyToKeyProto (C:\www\senfab\node_modules\@google-cloud\datastore\build\src\entity.js:974:19)
    at C:\www\senfab\node_modules\@google-cloud\datastore\build\src\entity.js:1108:45
    at Array.map (<anonymous>)
    at Object.queryToQueryProto (C:\www\senfab\node_modules\@google-cloud\datastore\build\src\entity.js:1104:43)
    at makeRequest (C:\www\senfab\node_modules\@google-cloud\datastore\build\src\request.js:273:49)
```

#### Results explanation

=> In build/src/entity.js l.1107, in the queryToQueryProto function, these lines are causing me trouble:
````js
if (filter.name === '__key__') {
    value.keyValue = entity.keyToKeyProto(filter.val);
}
else {
    value = entity.encodeValue(filter.val, filter.name);
}
````
=> `entity.encodeValue` handle correctly if `filter.val` is an array, but `entity.keyToKeyProto` doesn't handle arrays,  
whereas the API does handle filtering on `__key__` with the "IN" operator and an array as value.

#### Code suggestion

Patching this code is enough to make it work:
```js
if (filter.name === '__key__') {
  if (Array.isArray(filter.val)) {
    value = {
      arrayValue: {
        values: filter.val.map(v => {
          return {
            keyValue: entity.keyToKeyProto(v)
          }
        })
      }
    }
  } else {
    value.keyValue = entity.keyToKeyProto(filter.val)
  }
} else {
  value = entity.encodeValue(filter.val, filter.name)
}
```

#### Environment details

  - OS: Windows 10
  - Node.js version: 16.14.0
  - npm version: 8.3.1

malt03:

Fixes #1024 🦕

### DIFF
--- a/test/entity.ts
+++ b/test/entity.ts
@@ -1847,6 +1847,30 @@ describe('entity', () => {
             },
             {
               propertyFilter: {
+                op: 'IN',
+                property: {
+                  name: '__key__',
+                },
+                value: {
+                  arrayValue: {
+                    values: [
+                      {
+                        keyValue: {
+                          path: [
+                            {
+                              kind: 'Kind1',
+                              name: 'key1',
+                            },
+                          ],
+                        },
+                      },
+                    ],
+                  },
+                },
+              },
+            },
+            {
+              propertyFilter: {
                 property: {
                   name: '__key__',
                 },
@@ -1879,6 +1903,7 @@ describe('entity', () => {
       const query = ds
         .createQuery('Kind1')
         .filter('name', 'John')
+        .filter('__key__', 'IN', [new entity.Key({path: ['Kind1', 'key1']})])
         .start('start')
         .end('end')
         .groupBy(['name'])
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -1253,26 +1253,15 @@ export namespace entity {
     }
 
     if (query.filters.length > 0) {
-      const filters = query.filters.map(filter => {
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        let value: any = {};
-
-        if (filter.name === '__key__') {
-          value.keyValue = entity.keyToKeyProto(filter.val);
-        } else {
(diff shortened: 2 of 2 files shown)


test/entity.ts
```diff
@@ -1879,6 +1903,7 @@ describe('entity', () => {
       const query = ds
         .createQuery('Kind1')
         .filter('name', 'John')
+        .filter('__key__', 'IN', [new entity.Key({path: ['Kind1', 'key1']})])
```
danieljbruce, line 1906: We should make a separate test case to test this functionality. Especially since the new tests added will not include this filter so can't be used against this proto.

danieljbruce:
A similar change was adopted in https://github.com/googleapis/nodejs-datastore/pull/1085/files. Closing as this PR is now obselete.

PR #1067, opened 2023-02-15
fix: Allow filtering null values

### LINKED ISSUE #958: Unable to filter property with null value
#### Environment details

  - OS: MacOS
  - Node.js version: 16.15.0
  - npm version: 8.5.5

#### Steps to reproduce

Not sure if I'm missing something but currently any attempt to query by a value that might be null will result in TS error `Type 'null' is not assignable to type '{}'.`
e.g.
```
  const queryByStatus = (status: string | null) => {
    // Without operator
    const query1 = datastore.createQuery(kind).filter("status", status);
    // With operator
    const query2 = datastore.createQuery(kind).filter("status", "=", status);
  };
```

This is because the typings for `Query.filter` do not allow passing a null value:

```
filter(property: string, value: {}): Query;
filter(property: string, operator: Operator, value: {}): Query;
```

Note the workaround is to use the non-null operator `!` on the value but this isn't ideal.

danieljbruce:
A user expressed interest in filtering with a null value. Currently this is possible if a null value is provided in a variable, but if a `null` literal is passed into `value` for `.filter` then we get a typescript compiler error. This typescript parameter change will now allow users to not experience the typescript compiler error when trying to filter by null.

Fixes #958

### DIFF
--- a/system-test/datastore.ts
+++ b/system-test/datastore.ts
@@ -869,8 +869,19 @@ describe('Datastore', () => {
     });
 
     it('should construct filters by null status', async () => {
-      datastore.createQuery('Character').filter('status', null);
-      datastore.createQuery('Character').filter('status', '=', null);
+      const filtersNoEqual = datastore
+        .createQuery('Character')
+        .filter('status', null)
+        .filters.pop();
+      assert.strictEqual(filtersNoEqual ? filtersNoEqual.val : undefined, null);
+      const filtersWithEqual = datastore
+        .createQuery('Character')
+        .filter('status', '=', null)
+        .filters.pop();
+      assert.strictEqual(
+        filtersWithEqual ? filtersWithEqual.val : undefined,
+        null
+      );
     });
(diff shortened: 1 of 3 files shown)


kolea2 (commented):
LGTM, but please:

1. rename the PR to be clear this is for filtering (maybe something like 'fix: allow filtering null values')
2. Please have someone from yoshi-nodejs take a look

system-test/datastore.ts
```diff
@@ -868,6 +868,10 @@ describe('Datastore', () => {
       assert.strictEqual(entities.length, characters.length);
     });
 
+    it('should construct filters by null status', async () => {
+      datastore.createQuery('Character').filter('status', null);
+      datastore.createQuery('Character').filter('status', '=', null);
```
ruyadorno, line 873: It would be better to assert the returning result of this filter operation in all tests.

ruyadorno (approved):
LGTM 👍

PR #1075, opened 2023-03-01
fix: Or filters with warnings

danieljbruce:
- [ ] Ensure the tests and linter pass

Should be merged after or-queries PR. Adds warnings when `.filter` is not used correctly and modifies tests to react to warnings correctly.


src/query.ts
```diff
@@ -218,6 +218,9 @@ class Query {
       this.entityFilters.push(propertyOrFilter);
       return this;
     } else {
+      process.emitWarning(
+        'Providing Filter objects is recommended when using .filter'
```
kolea2, line 222: nit - mention an example like a CompositeFilter

kolea2 (approved):
lgtm after nit

kolea2:
also, please rename PR with conventional commit prefix

PR #1079, opened 2023-03-10
Change to try to ignore __key__ as it is redundant

danieljbruce:
This change matches the change shown in another PR, but the test cases won’t allow for the new change. There must be some edge cases that haven’t been considered here.

### DIFF
--- a/src/entity.ts
+++ b/src/entity.ts
@@ -416,6 +416,10 @@ export namespace entity {
       });
     }
 
+    isKey() {
+      return true;
+    }
+
     /**
      * Access the `serialized` property for a library-compatible way to re-use a
      * key.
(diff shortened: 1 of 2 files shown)


Guillaume-Sensome:
This PR is linked to #1024.  
I didn't find the time to dive into tests to make it pass with this change.

Thanks to anyone doing it.

vicb:
The test failures tell us that 

https://github.com/googleapis/nodejs-datastore/blob/main/src/entity.ts#L672-L675

is never taken.

Did some debugging, the root cause is that the `Key` in tests in a copy (i.e. not the same as) the lib `Key` due to the the way it's imported in test.

https://github.com/vicb/nodejs-datastore/commit/184f0ede1a897a5ae18f6d3d93bad5d7cec995ce proves it

danieljbruce:
https://github.com/googleapis/nodejs-datastore/pull/1085/files was the change adopted to solve this problem.

PR #1080, opened 2023-03-14
1065 query filter on key with adjustment

danieljbruce:
This demonstrates that if we update key checking then the tests can pass.


Guillaume-Sensome:
This PR would resolve https://github.com/googleapis/nodejs-datastore/issues/1024.

vicb:
This PR shows what the problem is but it's not a proper fix

PR #1085, opened 2023-03-14
fix: allow IN queries on __key__

### LINKED ISSUE #1024: Query filter on __key__ using the "IN" operator

danieljbruce:
The fix contains:

- An update to the way that entities are encoded
- An update to tests so that an instanceOf check passes
- A cleanup of the PropertyFilter function
- An additional test to check for IN with keys

Fixes #1024


src/filter.ts
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import {Operator, Filter as IFilter} from './query';
-import {entity} from './entity';
+import {entity, ValueProto} from './entity';
```
vicb, line 16: Please fix (remove `ValueProto`)

src/filter.ts
```diff
@@ -100,11 +81,7 @@ export class PropertyFilter extends EntityFilter implements IFilter {
    */
   // eslint-disable-next-line
   toProto(): any {
-    const value = new PropertyFilter(
-      this.name,
-      this.op,
-      this.val
-    ).encodedValue();
+    const value = entity.encodeValue(this.val, this.name);
```
vicb, line 84: You can also `value: entity.encodeValue(this.val, this.name)` inline. Not sure what the preferred style is.

vicb (approved):
Looks nice 👍 

There are a few "GitHub Actions / lint" that look worth addressing - IMO the typings of this lib could use some love but out of scope for this PR.

test/entity.ts
```diff
@@ -1894,8 +1894,62 @@ describe('entity', () => {
       assert.deepStrictEqual(entity.queryToQueryProto(query), queryProto);
     });
 
+    it('should support using key with IN', () => {
```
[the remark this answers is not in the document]
danieljbruce: Good point.

I think more specifically, it is addressing the case of using an array as a value as that was what was causing the error in the first place.

src/filter.ts
danieljbruce: Yes. This can be even more compact.

src/filter.ts
danieljbruce: Done

danieljbruce:
@vicb There is just one more thing we should confirm. I notice that `val` in Property Filter is of type `any` so even though filtering by `__key__` ought to use only values that are keys and arrays of keys, the compiler doesn't guarantee it. Are these definitely the only two cases with which we can expect users to filter with `__key__`?

src/filter.ts
```diff
@@ -56,42 +57,34 @@ export abstract class EntityFilter {
   abstract toProto(): any;
 }
 
+export type restrictedValueType<T> = T extends '__key__'
```
vicb, line 60: `RestrictedValueType`

src/query.ts
```diff
@@ -207,12 +207,19 @@ class Query {
...
-  filter(propertyOrFilter: string | EntityFilter, value?: {} | null): Query;
-  filter(propertyOrFilter: string, operator: Operator, value: {} | null): Query;
-  filter(
-    propertyOrFilter: string | EntityFilter,
+  filter<T extends string>(
+    propertyOrFilter: T | EntityFilter,
+    value?: restrictedValueType<T> | null
+  ): Query;
+  filter<T extends string>(
+    propertyOrFilter: T,
```
vicb, line 215: T | EntityFilter

src/query.ts
```diff
@@ -207,12 +207,19 @@ class Query {
...
+    operator: Operator,
+    value: restrictedValueType<T> | null
+  ): Query;
+  filter<T extends string>(
+    propertyOrFilter: T | EntityFilter,
     operatorOrValue?: Operator,
```
vicb, line 221: This should be `restrictedValueType<T> | null | Operator`

danieljbruce:
I mean see https://github.com/googleapis/nodejs-datastore/pull/1089/commits/8d269d2e2cd60bef27ff6856bffe21c5a8f9f20c

danieljbruce:
I mean https://github.com/googleapis/nodejs-datastore/pull/1089

src/query.ts
danieljbruce: Signature has been updated

src/filter.ts
```diff
@@ -156,5 +144,5 @@ class CompositeFilter extends EntityFilter {
 }
 
 export function isFilter(filter: any): filter is EntityFilter {
-  return (filter as EntityFilter).toProto !== undefined;
+  return filter instanceof EntityFilter;
```
kolea2, line 147: why does this have to change?

vicb: I suggested the fix so I'll answer.

`filter.toProto !== undefined` only checks if the object has a `toProto` properties (not set to `undefined`)
It is the case that `EntityFilter`s have a `toProto` methods but it is not necessarily the case that an object with a `toProto` property is an `EntityFilter`.

`filter instanceof EntityFilter` is the correct check for `filter is EntityFilter`

I hope that makes sense.

src/query.ts
```diff
@@ -207,12 +207,20 @@ class Query {
    * const keyQuery = query.filter('__key__', key);
    * ```
    */
-  filter(propertyOrFilter: string | EntityFilter, value?: {} | null): Query;
-  filter(propertyOrFilter: string, operator: Operator, value: {} | null): Query;
-  filter(
-    propertyOrFilter: string | EntityFilter,
-    operatorOrValue?: Operator,
-    value?: {} | null
+  filter(filter: EntityFilter): Query;
```
danieljbruce, line 210: This signature change makes the input parameters more restrictive. Do we really want to go through with this? If so would this be a major version bump?

@vicb @bcoe

vicb: My take is that the old signature was buggy and wouldn't catch unsupported calls. So I see this as a bug fix.

Also I think the impl could be improved by `switch`ing on `arguments.length`. 1, 2, 3 corresponding to the first 3 signatures in order.

danieljbruce: I will do this in a separate issue:

https://github.com/googleapis/nodejs-datastore/issues/1096

ruyadorno (approved):
LGTM 👍 

Let's just make sure to keep an eye and monitor if the internal change from `src/filter.ts` `encodeValue()` is going to have any impact in the end users of the library. Releasing this as a patch version release is going to help us having the flexibility to rollback if needed, so I'd advise to proceed with a `fix:` commit.

Great work @danieljbruce!

PR #1097, opened 2023-03-27
feat: Sum and average aggregation queries

danieljbruce:
This PR demonstrates how we are going to leverage the current framework to support sum/average queries for later use.


src/aggregate.ts
```diff
@@ -157,4 +196,58 @@ class Count extends AggregateField {
...
+    super();
+    this.property_ = property;
+  }
+
+  /**
+   * Gets the proto for the property aggregate field.
+   *
+   */
+  toProto(): any {
+    const aggregation = Object.assign(
```
vicb, line 223: Could you just
```js
const aggregation = this.property_ ? {property: {name: this.property_}} : {};
```

src/aggregate.ts
```diff
@@ -157,4 +196,58 @@ class Count extends AggregateField {
...
+      {},
+      this.property_ ? {property: {name: this.property_}} : null
+    );
+    const aggregationObject: any = {};
+    aggregationObject[this.operator] = aggregation;
+    return Object.assign(
+      {operator: this.operator},
+      this.alias_ ? {alias: this.alias_} : null,
+      aggregationObject
```
vicb, line 232: You can also drop the intermediate object and replace this line by
```js
{[this.operator]: aggregation}
```

src/aggregate.ts
```diff
@@ -157,4 +196,58 @@ class Count extends AggregateField {
...
+abstract class PropertyAggregateField extends AggregateField {
+  property_?: string;
+  abstract operator: string;
+
+  /**
+   * Build a PropertyAggregateField object.
+   *
+   * @param {string} property
+   */
+  constructor(property: string) {
```
vicb, line 213: nit: 
```js
constructor(public property_: string) {
```

and remove the fields + the assignment

src/aggregate.ts
danieljbruce: done

src/aggregate.ts
danieljbruce: done

src/aggregate.ts
danieljbruce: done

system-test/datastore.ts
```diff
@@ -1130,6 +1369,110 @@ describe('Datastore', () => {
...
+    after(async () => {
+      await datastore.delete(keys);
+    });
+    it('should run a sum aggregation', async () => {
+      const q = datastore.createQuery('Character');
+      const aggregate = datastore
+        .createAggregationQuery(q)
+        .addAggregation(AggregateField.sum('appearances'));
+      const [results] = await datastore.runAggregationQuery(aggregate);
+      assert.deepStrictEqual(results, [{property_1: -18446744073709552000}]);
```
danieljbruce, line 1414: I can confirm that this value and the value for average are the same values received by the grpc call so the values you see here are strictly backend calculations.

src/request.ts
```diff
@@ -891,13 +891,26 @@ class DatastoreRequest {
...
+    if (options.readTime) {
+      if (sharedQueryOpts.readOptions === undefined) {
+        sharedQueryOpts.readOptions = {};
+      }
+      const readTime = options.readTime;
+      const seconds = readTime / 1000;
+      sharedQueryOpts.readOptions.readTime = {
+        seconds: Math.floor(seconds),
+      };
+    }
```
danieljbruce, line 913: This code was taken from https://github.com/googleapis/nodejs-datastore/blob/e7c3aaf665ebae5a91fb692bb0af54717e5ea587/src/request.ts#L290-L299 and can be refactored after the merge as per the TODO.

src/request.ts
```diff
@@ -891,13 +891,26 @@ class DatastoreRequest {
     query: Query,
     options: RunQueryStreamOptions = {}
   ): SharedQueryOptions {
+    // TODO: This needs to be refactored later:
```
kolea2, line 894: is this to be done in this PR or in a future one?

src/request.ts
kolea2: any reason why we can't refactor this now?

system-test/datastore.ts
```diff
@@ -704,6 +710,13 @@ describe('Datastore', () => {
       assert.strictEqual(secondEntities!.length, 3);
     });
 
+    it('should query the datastore with snapshot read', async () => {
```
kolea2, line 713: is this related to SUM/AVG changes? If not, let's move it into a separate PR

system-test/datastore.ts
```diff
@@ -950,6 +963,235 @@ describe('Datastore', () => {
...
+          .addAggregation(AggregateField.sum('appearances'));
+        const [results] = await datastore.runAggregationQuery(aggregate);
+        assert.deepStrictEqual(results, [{property_1: 187}]);
+      });
+      it('should run a sum aggregation with a list of aggregates', async () => {
+        const q = datastore.createQuery('Character');
+        const aggregate = datastore
+          .createAggregationQuery(q)
+          .addAggregations([
+            AggregateField.sum('appearances'),
```
kolea2, line 980: might be more interesting to include a sum, average, and a count all in one query

system-test/datastore.ts
```diff
@@ -1130,6 +1372,110 @@ describe('Datastore', () => {
...
+          key,
+          data: characters[index],
+        };
+      });
+      await datastore.save(keysToSave);
+    });
+    after(async () => {
+      await datastore.delete(keys);
+    });
+    it('should run a sum aggregation', async () => {
```
danieljbruce, line 1411: Include `overflow` word here

src/request.ts
danieljbruce: No reason. It is a choice between having two simpler PRs and one PR that includes both changes. Doing so now is fine though.

src/request.ts
danieljbruce: Same the above comment. We can do it now if preferred.

system-test/datastore.ts
danieljbruce: ok

system-test/datastore.ts
danieljbruce: I could do that in a separate block. This test is inside the `with a sum filter` block.

danieljbruce: Done

system-test/datastore.ts
danieljbruce: Done

src/aggregate.ts
```diff
@@ -121,6 +144,24 @@ abstract class AggregateField {
...
+    return new Sum(property);
+  }
+
+  /**
+   * Gets a copy of the Average aggregate field.
+   *
+   * @returns {Average}
+   */
+  static average(property: string): Sum {
+    return new Average(property);
```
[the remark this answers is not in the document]
kolea2: this looks done, good to resolve?

src/aggregate.ts
```diff
@@ -121,14 +144,32 @@ abstract class AggregateField {
...
+   * Gets a copy of the Average aggregate field.
+   *
+   * @returns {Average}
+   */
+  static average(property: string): Average {
+    return new Average(property);
+  }
+
   /**
    * Gets a copy of the Count aggregate field.
```
kolea2, line 166: should this be updated to remove "Count" since it applies to all?

src/aggregate.ts
```diff
@@ -121,14 +144,32 @@ abstract class AggregateField {
...
    *
    * @param {string} alias The label used in the results to describe this
    * aggregate field when a query is run.
    * @returns {AggregateField}
    */
-  alias(alias: string): AggregateField {
+  alias(alias?: string): AggregateField {
```
kolea2, line 172: can we add some additional testing on Count now that this is optional?

src/request.ts
```diff
@@ -887,9 +867,8 @@ class DatastoreRequest {
     return stream;
   }
 
-  private getSharedQueryOptions(
-    query: Query,
-    options: RunQueryStreamOptions = {}
+  private getSharedOptionsOnly(
```
kolea2, line 870: what is the difference between this method and `getSharedQueryOptions`? The naming is confusing here.

system-test/datastore.ts
```diff
@@ -676,12 +677,17 @@ describe('Datastore', () => {
     ];
 
     before(async () => {
+      function sleep(ms: number) {
```
kolea2, line 680: why are we adding this? Can you add a comment above explaining?

system-test/datastore.ts
```diff
@@ -1014,6 +1249,22 @@ describe('Datastore', () => {
         assert.deepStrictEqual(results, [{total: 7}]);
       });
     });
+    describe('with multiple types of filters', () => {
+      it('should run multiple types of aggregations with a list of aggregates', async () => {
+        const q = datastore.createQuery('Character');
+        const aggregate = datastore
+          .createAggregationQuery(q)
+          .addAggregations([
+            AggregateField.count(),
```
kolea2, line 1258: I would add one more test here using a combination of aliases and non aliases (aka count with an alias, sum without an alias, avg with an alias all in one query)

kolea2 (changes requested):

src/aggregate.ts
danieljbruce: Done

src/request.ts
danieljbruce: `getSharedQueryOptions` also accepts a query and calls `getSharedOptionsOnly`. `getSharedOptionsOnly` is also called for fetches in `createReadStream` where options need to be built without considering a query.

src/aggregate.ts
danieljbruce: Yes. Added in commit called "Added tests for when an empty alias is provided".

system-test/datastore.ts
danieljbruce: Done

danieljbruce: https://github.com/googleapis/nodejs-datastore/pull/1097/commits/bbff19ff13c0eabb19dee64536b2e493d5982cf9

src/request.ts
kolea2: Thanks for the explanation! Maybe `getSharedQueryOptions` and `getSharedRequestOptions`, or `getQueryOptions` and `getRequestOptions` would be a little more readable, wdyt?

danieljbruce: `getQueryOptions` and `getRequestOptions` would work really well. The good thing about these two functions is that they are both private so we can change the names later too if need be.

system-test/datastore.ts
```diff
@@ -676,12 +677,18 @@ describe('Datastore', () => {
     ];
 
     before(async () => {
+      // This function is used for testing snapshot reads by getting a time and then sleeping before adding data.
```
ruyadorno, line 680: I believe it would be important to rephrase this comment so that it explains **why** we need the sleeping: 
is it simulating process time from datastore? is it simulating latency?

ruyadorno (commented):
A couple of questions and tweaking a comment but otherwise it's looking good to me 👍

src/index.ts
```diff
@@ -40,8 +40,9 @@ import * as is from 'is';
 import {Transform, pipeline} from 'stream';
 
 import {entity, Entities, Entity, EntityProto, ValueProto} from './entity';
+import {AggregateField} from './aggregate';
 import Key = entity.Key;
-export {Entity, Key};
```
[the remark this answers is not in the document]
danieljbruce: See design doc comment.

src/index.ts
```diff
@@ -1818,7 +1819,6 @@ promisifyAll(Datastore, {
     'isDouble',
     'geoPoint',
     'getProjectId',
-    'getSharedQueryOptions',
```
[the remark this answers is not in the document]
ruyadorno: makes sense then, so it sounds like `getSharedQueryOptions` was added here in error since it also lived in `request.ts`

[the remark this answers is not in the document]
danieljbruce: Yes. I think so.

system-test/datastore.ts
```diff
@@ -689,8 +709,16 @@ describe('Datastore', () => {
           data: characters[index],
         };
       });
-      timeBeforeDataCreation = Date.now();
-      await sleep(1000);
+      // Save for a key so that a read time can be accessed for snapshot reads.
+      const dummyData = Object.assign(Object.assign({}, keysToSave[0]), {
```
kolea2, line 713: rename to "emptyData"

kolea2 (changes requested):

system-test/datastore.ts
danieljbruce: Done

kolea2 (approved):
lgtm, please wait for @ruyadorno's approval as well before merging. Thanks for all the hard work on this!

ruyadorno (commented):
LGTM 👍

PR #1099, opened 2023-03-30
test: Remove mock so that future tests will not be affected by the mock

danieljbruce:
This issue was discovered when trying to wrap `test/index` in async. Tests were failing because a mock in one test was being used again in future tests and caused those tests to fail in the second async loop. The change involves resetting the mock to its original setup so that future tests do not use the mock that was set up in that one particular test.

### DIFF
--- a/test/index.ts
+++ b/test/index.ts
@@ -26,6 +26,8 @@ import * as is from 'is';
 import * as sinon from 'sinon';
 import * as extend from 'extend';
 
+const ENTITY_TO_ENTITY_PROTO = entity.entityToEntityProto;
+
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const v1 = require('../src/v1/index.js');
 
@@ -77,7 +79,7 @@ const fakeEntity: any = {
   keyFromKeyProto: entity.keyFromKeyProto,
   keyToKeyProto: entity.keyToKeyProto,
   encodeValue: entity.encodeValue,
-  entityToEntityProto: entity.entityToEntityProto,
+  entityToEntityProto: ENTITY_TO_ENTITY_PROTO,
   findLargeProperties_: entity.findLargeProperties_,
   URLSafeKey: entity.URLSafeKey,
 };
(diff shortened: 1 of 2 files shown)


test/index.ts
```diff
@@ -1774,6 +1776,8 @@ describe('Datastore', () => {
           (config.reqOpts!.mutations![0].upsert! as Entity).excludeFromIndexes,
           excludeFromIndexes
         );
+        // Reassign this to what it was so that other tests don't fail because of the mock.
+        fakeEntity.entityToEntityProto = ENTITY_TO_ENTITY_PROTO;
```
kolea2, line 1780: maybe surround this test with a try/finally block to ensure this resets even if the test fails?

kolea2 (changes requested):

test/index.ts
danieljbruce: Done

danieljbruce: Reverted. Others had comments so may chime in a bit later.

vicb: I don't think `try`/`finally` would be a good fit here because it's sync and the test is async (that is why might reset the prop before it is actually used).

I am looked at this super closely but there are several things to explore to reset the prop IMO:
- use an `async` test fn if the test framework and the code flow support it,
- use `afterTest` + global objet to restore explicitly saved props (that would look a little hacky)

danieljbruce: By the last comment do you mean wrapping this test in a describe block and then adding an after hook in the describe block?

vicb: Actually I don't think my first option would work.

Maybe something to explore:

```js
const fakeEntityInit: any = {
 // previous fakeEntity
};

const fakeEntity: any = {};

// ...

  beforeEach(() => {
    // init fakeEntity
    fakeEntity[KEY_SYMBOL] = fakeEntityInit[KEY_SYMBOL];
    for (const [key, value] of Object.entries(fakeEntityInit)) {
      fakeEntity[key] = value;
    }

    // existing code   
    createInsecureOverride = null;
    googleAuthOverride = null;
    datastore = new Datastore({
      projectId: PROJECT_ID,
      namespace: NAMESPACE,
    });
  });
```

This would ensure the `fakeEntity` is init from a clean value before each test.

danieljbruce: This would work. I guess we don't need to care too much about the extra run time this would add since these are tests.

danieljbruce: I adopted the same idea, but did it a slightly different way. We do not want to have to iterate through each property like this. We can just use `Object.assign` and that will do the trick.

kolea2 (approved):

PR #1101, opened 2023-04-06
fix: Allow user to set custom endpoints

danieljbruce:
This change allows the user to set a custom endpoint when creating the client and it will work with the backend the way it is supposed to. It also ensures that we can still work with the emulator using the client.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -507,7 +507,8 @@ class Datastore extends DatastoreRequest {
     const isUsingEmulator =
       this.baseUrl_ &&
       (this.baseUrl_.includes('localhost') ||
-        this.baseUrl_.includes('127.0.0.1'));
+        this.baseUrl_.includes('127.0.0.1') ||
+        this.baseUrl_.includes('::1'));
     if (this.customEndpoint_ && isUsingEmulator) {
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -504,8 +504,12 @@ class Datastore extends DatastoreRequest {
       },
       options
     );
-    if (this.customEndpoint_) {
-      this.options.sslCreds = grpc.credentials.createInsecure();
+    const isUsingEmulator =
+      this.baseUrl_ &&
+      (this.baseUrl_.includes('localhost') ||
+        this.baseUrl_.includes('127.0.0.1'));
```
alexander-fenster, line 510: Just in case, maybe add `::1` as an IPv6 version of localhost?

alexander-fenster (approved):

src/index.ts
danieljbruce: Done

edance:
hey all! This PR broke the docker setup for us. We're running the datastore emulator using a docker container so the datastore emulator host is something like `datastore:8081`. We want a connection that is both insecure and not to localhost.

How can we make changes to this code? If `process.env.DATASTORE_EMULATOR_HOST` is set, should we make `isUsingEmulator == true`

PR #1103, opened 2023-04-15
chore: de-duplicate sample region tags

meredithslota:
Testing removing these samples from the combined `tasks.js` file as they exist in separate files. Do not merge just yet, other updates may be needed.


meredithslota:
Ok, just kidding — this requires other updates as the samples for Datastore are structured a bit differently, and snippet-bot isn't catching this. Closing.

PR #1108, opened 2023-05-03
chore(filter): Add a switch statement

### LINKED ISSUE #1096: Change filter in query implementation to use a switch
As mentioned in the following pull request:

https://github.com/googleapis/nodejs-datastore/pull/1085

We should `switch` on `arguments.length` in the `filter` function of `query`.

danieljbruce:
In the filter function a switch statement is added so that it is clear to the reader what happens when the arguments are of various lengths.

Fixes #1096

### DIFF
--- a/src/query.ts
+++ b/src/query.ts
@@ -231,19 +231,13 @@ class Query {
       case 1: {
         if (isFilter(propertyOrFilter)) {
           this.entityFilters.push(propertyOrFilter);
-        } else {
-          this.filters.push({
-            name: (propertyOrFilter as String).trim(),
-            op: (operatorOrValue as Operator).trim() as Operator,
-            val: value,
-          });
         }
         break;
       }
       case 2: {
         this.filters.push({
           name: (propertyOrFilter as String).trim(),
-          op: '=' as Operator,
+          op: '=',
           val: operatorOrValue as AllowedFilterValueType<T>,
         });
         break;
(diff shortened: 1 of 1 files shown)


src/query.ts
```diff
@@ -222,24 +222,39 @@ class Query {
...
-      let operator = operatorOrValue as Operator;
-      if (arguments.length === 2) {
-        value = operatorOrValue as AllowedFilterValueType<T>;
-        operator = '=';
+    }
+    switch (arguments.length) {
+      case 1: {
+        if (isFilter(propertyOrFilter)) {
+          this.entityFilters.push(propertyOrFilter);
+        } else {
```
vicb, line 234: I think the `else` branch should be removed here (`operatorOrValue` and `value` are not provided)

src/query.ts
```diff
@@ -222,24 +222,39 @@ class Query {
...
+            op: (operatorOrValue as Operator).trim() as Operator,
+            val: value,
+          });
+        }
+        break;
+      }
+      case 2: {
+        this.filters.push({
+          name: (propertyOrFilter as String).trim(),
+          op: '=' as Operator,
```
vicb, line 246: `as Operator` is probably not needed here?

vicb (approved):
LGTM with a couple nits

src/query.ts
danieljbruce: ok

src/query.ts
danieljbruce: yup

vicb (approved):
Nice!

ruyadorno (approved):

PR #1114, opened 2023-05-26
fix: Check property existence for exclude from indexes with wildcard

### LINKED ISSUE #787: excludeFromIndexes using wildcard  on non-existing data causes `TypeError: Cannot read property 'entityValue' of undefined`.

danieljbruce:
Changes:

Adds a check in three places for the existence of a property in the path provided for exclude from indexes. Also adds test cases to explore what happens when a property is not contained in exclude from indexes, when the property is contained and when it is contained with an array in the path.

Take a look at the changes in the `src` folder. All the changes in the PR involve adding an additional check to see if `entity.properties![firstPathPart]` is defined before accessing a property of `entity.properties![firstPathPart]`. That solves the issue users are seeing where the code throws an error. Instead it just doesn't explore the non-existent property path further and it doesn't add `excludeFromIndexes = true` to any of the entity properties.

The exact test case corresponding to the issue the user reported is at https://github.com/googleapis/nodejs-datastore/pull/1114/files#diff-177d97aec07fd5b22f04049b34d74939b1d8364a7a83e8fb69ee3ee8695105fbR2187-R2198, but similar test cases where the same error would occur were added and source code changes were made to address those test cases.

Fixes #787


src/entity.ts
```diff
@@ -855,6 +855,7 @@ export namespace entity {
       if (
         firstPathPartIsArray &&
         // check also if the property in question is actually an array value.
+        typeof entity.properties![firstPathPart] !== 'undefined' &&
```
ruyadorno, line 858: curious: why the check used is `typeof entity.properties![firstPathPart] !== 'undefined'` and not `entity.properties![firstPathPart] !== undefined` ?

src/entity.ts
```diff
@@ -898,22 +904,26 @@ export namespace entity {
...
-
-            if (parentEntity) {
-              Object.keys(parentEntity.properties).forEach(path => {
-                const newPath = parentEntity.properties[path].arrayValue
-                  ? path + '[].*'
-                  : path + '.*';
-                excludePathFromEntity(parentEntity, newPath);
-              });
+          if (typeof entity.properties![firstPathPart] !== 'undefined') {
+            if (hasWildCard && remainderPath === '*') {
```
ruyadorno, line 908: I believe this can be simplified to:

```suggestion
          if (typeof entity.properties![firstPathPart] !== 'undefined' && hasWildCard && remainderPath === '*') {
```

src/entity.ts
danieljbruce: This change causes the test called `should encode a request when using a wildcard` to fail with `TypeError: Cannot read properties of undefined (reading 'arrayValue')`. I guess `typeof entity.properties![firstPathPart]` must evaluate to a string.

src/entity.ts
danieljbruce: The problem is that the latter if has an else statement.

src/entity.ts
danieljbruce: Nevermind, I forgot to remove typeof as per your comment. This change is fine.

danieljbruce: I changed this everywhere and refactored this change out into a variable.

src/entity.ts
danieljbruce: I added this change an changed the else to an if else so that the indenting doesn't make it so difficult to read.

src/entity.ts
ruyadorno: looks great 👍

test/index.ts
```diff
@@ -2150,4 +2160,125 @@ describe('Datastore', () => {
...
+
+    async.each(
+      onSaveTests,
+      (onSaveTest: {
+        properties: google.datastore.v1.IValue;
+        entitiesWithoutKey: Entities;
+      }) => {
+        it(`should pass the right properties to upsert on save with parameters: ${JSON.stringify(
+          onSaveTest
+        )}`, async () => {
```
ruyadorno, line 2231: What if we added a `description` with a human-readable, clear description of each scenario (it's probably fine to just reuse the wording for the comments on top of each object definition) instead of just dropping the blob of `JSON.stringify()` which I'm afraid will get really unreadable and harder to maintain with time.

danieljbruce: Sounds good. Then we can use the description as the test name.

danieljbruce: Done

ruyadorno (approved):
LGTM 👍

PR #1144, opened 2023-08-30
test: Fix logger for export import test

### LINKED ISSUE #1143: The import/export test should delay before retrying and should emit info logs
An [import/export test](https://github.com/googleapis/nodejs-datastore/blob/e7c3aaf665ebae5a91fb692bb0af54717e5ea587/system-test/datastore.ts#L1337) which is supposed to [emit console infos](https://github.com/googleapis/nodejs-datastore/blob/e7c3aaf665ebae5a91fb692bb0af54717e5ea587/system-test/datastore.ts#L1332) and allow delays does not actually do so and therefore the test doesn't function as intended.

[currentRetry](https://github.com/googleapis/nodejs-datastore/blob/e7c3aaf665ebae5a91fb692bb0af54717e5ea587/system-test/datastore.ts#L1327) does not actually correspond to a function and so we must introduce a variable that lives on the test context that tracks the number of attempts in the test and use that instead.

In the test's current state, delay tries to call currentRetry, but fails silently because delay is an async function. The code compiles fine because currentRetry can match with [this property](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/7e3ed2122891e449b3b42764ad693ecf637ce849/types/mocha/index.d.ts#L1372) which has type any.

The info logs will also help diagnose the root cause of https://github.com/googleapis/nodejs-datastore/issues/1122

danieljbruce:
The change involves the following:
- Adding a change to the delay function to use a variable on the mocha test context instead of the currentRetry function which currently does not correspond to a function that lives on the context
- Adding a setup function which sets up this variable each time any test is run that uses the delay function
- Adding a test for these two functions to document and ensure they behave in such a way that console.info is actually called and the test is retried the appropriate number of times.

Fixes [#1143 ](https://github.com/googleapis/nodejs-datastore/issues/1143)

The console.info logs will further help diagnose flakey tests in https://github.com/googleapis/nodejs-datastore/issues/1122


system-test/datastore.ts
```diff
@@ -1323,18 +1323,61 @@ describe('Datastore', () => {
...
+      let consoleInfoFunction: (message: string) => void;
+      let infoLogCount = 0;
+
+      before(async () => {
+        consoleInfoFunction = console.info;
+      });
+
+      it('should be sure that the delay function emits console info messages', async function () {
+        // Override console.info to track the number of times it is called.
+        console.info = consoleInfoFunction;
```
bhshkh, line 1353: Why is this required if the next line is overriding this assignment?

danieljbruce: I think this was a typo which needs to be removed.

bhshkh (approved):

PR #1172, opened 2023-10-16
test: Change the filter samples to use entity filters

danieljbruce:
Entity filters should be used in most places instead of the legacy filter functionality. This PR ensures that the entity filters are used in all the code samples. This discourages users from using the filter function the old way as we want them to realize the benefits of using the new entity filters.


daniel-sanche (approved):
LGTM

bhshkh (approved):

Issue #1109, opened 2023-05-04
Do not clutter the logs with warnings in prod environment


vicb:
The [following code](https://github.com/googleapis/nodejs-datastore/blob/cd59c718ebe6711603b08b3ba2677e5a984bc7e3/src/query.ts#LL198C1-L200C34)

```js
 const companyQuery = query
   .filter('name', 'Google')
   .filter('size', '<', 400);
```

will print a warning:

```js
process.emitWarning(
  'Providing Filter objects like Composite Filter or Property Filter is recommended when using .filter'
);
```

First I don't think there should be any warning in a prod environment.
One way js libraries check for prod environemnt is to check if the `NODE_ENV` environment variable is set to `production` - see the [express docs](https://expressjs.com/en/advanced/best-practice-performance.html#set-node_env-to-production) for ref.

Also my understanding is that the warning is there because of nested composite filters. I don't think there should be any warnings in a simple case, that is:

```js
const companyQuery = query.filter('name', 'Google');
```

should probably not generate a warning even in dev env.

/cc @danieljbruce

danieljbruce:
[Datastore in Python](https://github.com/googleapis/python-datastore/blob/b31a94455365445f5abc4eefc5edbef275ecc750/google/cloud/datastore/query.py#L432) emits this warning too so whatever we end up doing there should probably be consistency across the board.

vicb:
I am not familiar with Python `warnings` but it looks like you can at least [filter the warnings](https://docs.python.org/3/library/warnings.html#the-warnings-filter)

vicb:
This is what I mean:

<img width="1122" alt="image" src="https://github.com/googleapis/nodejs-datastore/assets/248818/f64a6440-93f8-4e3f-9acf-25a5efca242f">

It would be great to filter out all the noise in my prod env.

Thanks!

MiguiTE:
What should be done not to get this warning when building a query with filters?

According to docs there are two ways of filtering a query, [with](https://cloud.google.com/nodejs/docs/reference/datastore/latest/datastore/query#_google_cloud_datastore_Query_filter_member_2_) operator and [without](https://cloud.google.com/nodejs/docs/reference/datastore/latest/datastore/query#_google_cloud_datastore_Query_filter_member_1_) (defaults to equals). So, how could we build and supply a _Filter object_?

vicb:
It looks like the doc is not up to date. The filter method can now take objects. You can look at the tests for an example.

[...]

PR #1185, opened 2023-10-23
fix: Change tests to use new filter and use gax warn for warning just once

danieljbruce:
**Summary:**
To partially address https://github.com/googleapis/nodejs-datastore/issues/1109 we use a warning function internally that will emit a warning only when the legacy `.filter` function is used the first time. Right now, a warning is emitted every time the legacy filter function is used.

**Changes:**
**In src/query.ts:**
Changes are made inside the src/query.ts file to use the function that will limit warnings to 1.
**In tests outside of test/query.ts:**
.filter is called the new way by providing PropertyFilter and CompositeFilter objects, this ensures that when the test/query.ts file is run, a warning about using .filter the legacy way hasn't been emitted yet
**In test/query.ts:**
A test is moved up so that it is the first test that uses the legacy filter function. This test expects a warning so this change is necessary to make this test pass.

### DIFF
--- a/test/query.ts
+++ b/test/query.ts
@@ -206,6 +206,18 @@ describe('Query', () => {
   });
 
   describe('filter', () => {
+    it('should issue a warning when a Filter instance is not provided', done => {
+      const onWarning = (warning: {message: unknown}) => {
+        assert.strictEqual(
+          warning.message,
+          'Providing Filter objects like Composite Filter or Property Filter is recommended when using .filter'
+        );
+        process.removeListener('warning', onWarning);
+        done();
+      };
+      process.on('warning', onWarning);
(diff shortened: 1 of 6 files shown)


kolea2 (approved):

test/query.ts
```diff
@@ -206,6 +206,18 @@ describe('Query', () => {
   });
 
   describe('filter', () => {
+    it('should issue a warning when a Filter instance is not provided', done => {
```
feywind, line 209: It might be nice to make this validate that the warning only happens once, but since gax probably also tests that, it's probably not a big issue.

feywind (approved):

test/query.ts
danieljbruce: Done

PR #1196, opened 2023-11-02
refactor: Break transaction.run into smaller pieces for use with async functions and other read/write calls

danieljbruce:
Process summary:

Branch nodejs-transaction-redesign-feature-branch is going to eventually be merged into main and contain all work on redesigning the way transactions work in nodejs-datastore.

Summary of PR:

This PR is the first of many to make sure that when users are using the transaction object that they are always running read/write calls as a transaction. The end goal is that if a transaction has not been started yet then a request will be made to start the transaction before the read/write call is made.

To manage concurrency issues, a mutex will also be used. For commit/write and various transaction read functions, the business logic is complex and will be much easier to read and implement if it is written in an async function. If new commit and read function business logic is not in an async function then the functions will need callbacks for acquiring the mutex and running the begin transaction request which will be difficult to read/implement code. We want the function that begins the transaction to be async so that we can just await its result in the async functions for commit and the various read calls.

This PR just breaks apart the run function in the transaction object into smaller pieces and does the begin transaction call in an async function. This async function is going to be reused by commit and other read functions.

Changes:

Tests are added that capture how run currently works. They mock out beginTransaction in the gapic layer and capture what happens when an error or response was sent back from the gapic layer. This was necessary to be sure that when changes were made to transaction.run that run would behave exactly the same way as before.

The run function is broken up into 3 parts. run will now use two functions runAsync and parseRunAsync. runAsync is going to get used by the new functions for commit and get to begin the transaction. parseRunAsync is going to be used to parse the data returned from running the transaction (ie. saving the transaction id). runAsync and parseRunAsync just replace code that currently exists today for the run function.


src/transaction.ts
```diff
@@ -562,23 +596,27 @@ class Transaction extends DatastoreRequest {
...
+    ) => {
+      this.request_(
+        {
+          client: 'DatastoreClient',
+          method: 'beginTransaction',
+          reqOpts,
+          gaxOpts: options.gaxOptions,
+        },
+        // In original functionality sometimes a response is provided when an error is also provided
+        // reject only allows us to pass back an error so use resolve for both error and non-error cases.
```
daniel-sanche, line 610: What do you mean by original functionality here? Is this comment describing how it worked before the change?

src/transaction.ts
```diff
@@ -544,10 +556,32 @@ class Transaction extends DatastoreRequest {
...
+  // TODO: Replace with #parseRunAsync when pack and play error is gone
+  private parseRunAsync(
+    response: RequestPromiseReturnType,
+    callback: RunCallback
+  ): void {
+    const err = response.err;
+    const resp = response.resp;
+    if (err) {
+      callback(err, null, resp);
+      return;
```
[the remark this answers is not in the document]
danieljbruce: Done. This was copy / pasted from what was there previously in https://github.com/googleapis/nodejs-datastore/blob/3b2712b0e265c720d18c260e28f7239d7ebaf9d5/src/transaction.ts#L574-L579.

src/transaction.ts
```diff
@@ -544,10 +556,32 @@ class Transaction extends DatastoreRequest {
...
     const callback =
       typeof optionsOrCallback === 'function' ? optionsOrCallback : cb!;
+    this.runAsync(options).then((response: RequestPromiseReturnType) => {
+      this.parseRunAsync(response, callback);
+    });
+  }
 
-    const reqOpts = {
+  // TODO: Replace with #parseRunAsync when pack and play error is gone
+  private parseRunAsync(
```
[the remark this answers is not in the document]
danieljbruce: Changed the function name. Yes, this function is for processing the results from the begin transaction call. Added comments for the signature of each of these functions. runAsync should contain a lot of the same code that run used to have https://github.com/googleapis/nodejs-datastore/blob/3b2712b0e265c720d18c260e28f7239d7ebaf9d5/src/transaction.ts#L543-L581.

src/transaction.ts
daniel-sanche: Yeah, I'd say don't reference the way the code was before, make sure the comments explain the current state of the program

daniel-sanche (approved):
Mostly LGTM, but might be a good idea to have a node person review it as well

src/transaction.ts
danieljbruce: Sounds good. Comment has been changed.

src/transaction.ts
```diff
@@ -32,6 +32,18 @@ import {
 } from './request';
 import {AggregateQuery} from './aggregate';
 
+// RequestPromiseReturnType should line up with the types in RequestCallback
+interface RequestPromiseReturnType {
+  err?: Error | null;
+  resp: any; // TODO: Replace with google.datastore.v1.IBeginTransactionResponse and address downstream issues
```
ruyadorno, line 38: are you going to address this TODO comment before landing? what is the plan here?

ruyadorno (approved):
LGTM 👍

src/transaction.ts
danieljbruce: Not in this PR. The issue is that the id at https://github.com/googleapis/nodejs-datastore/blob/b86c79b920a8e014a6426332ec1819455e05694b/src/request.ts#L82 needs to include more types and changing the type for id to include Buffer for example is breaking. Let's discuss this further as a future change though because the types really should line up. In a future PR on this branch, this interface is going to be generalized.

I think when this feature branch collects the stacked PRs then we can do a final scan for TODOs. Does that sound ok?

PR #1202, opened 2023-11-17
refactor: Move commit logic and add tests that prepare for transaction function changes

danieljbruce:
**Summary:**

This pull request is a successor to https://github.com/googleapis/nodejs-datastore/pull/1196. In this pull request we just move some commit logic and add more testing to prepare for the next PR. The next PR will make meaningful changes to how transactions work in nodejs-datastore and we don't want trivial changes from this PR to show up there as well.

**Changes:**

**Changes in src directory**

All the functionality that used to live in the commit function now lives in the private #runCommit function. Now commit. doesn't do anything except call #runCommit. When a user calls transaction.commit then the same thing should happen as before. We are putting the functionality in #runCommit because in the next PR, transaction.commit will use another function that calls #runCommit and making this change now will make the diff easier to read in the next PR.

**Changes in test directory**

A suite of tests are added under the describe block _testing various transaction functions when transaction.run returns a response_. These tests make sure that in the upcoming PR the changes to various transaction functions will not introduce regression bugs. They are all tests that mock out the gapic layer and ensure the right return values are received by the user.

For the transaction.run tests added in the previous PR some small refactors are made. Some code to mock out beginTransaction is pulled into a function as it is used more than once. Try/catch blocks are also added inside a callback so that the error will bubble up to the test runner for easier debugging.

**Other changes**

A pull was done from main so changes in .github, .kokoro and package.json are showing up.


daniel-sanche (approved):
LGTM

ruyadorno (approved):

PR #1203, opened 2023-11-17
feat: Allow the user to set fallback to rest

danieljbruce:
**Change:**

Users need to be able to use the rest fallback property. This change allows client library calls to be rest calls instead of grpc calls. They can make this happen by instantiating the client library like so:

`new Datastore({namespace, fallback: 'rest'})`


test/gapic-mocks/client-initialization-testing.ts
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ * This class mocks out the lookup function so that for tests in this file
+ * the lookup function just sends data back instead of making a call to the
+ * server. The class also saves the rest parameter in the constructor so that
+ * it can be read later for correctness.
+ *
+ */
+class FakeDatastoreClient extends DatastoreClient {
+  restParameter: string | undefined;
```
cindy-peng, line 44: Hi Dan, just wondering is this "restParameter" expecting a boolean type from `Fallback type` too? https://github.com/googleapis/nodejs-datastore/pull/1203/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R100

danieljbruce: I changed the type to `Fallback` here so that the types line up better.

src/index.ts
```diff
@@ -1912,6 +1913,7 @@ export interface DatastoreOptions extends GoogleAuthOptions {
   apiEndpoint?: string;
   sslCreds?: ChannelCredentials;
   databaseId?: string;
+  fallback?: Fallback;
```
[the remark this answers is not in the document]
danieljbruce: I added a simple system test for this.

test/gapic-mocks/client-initialization-testing.ts
```diff
@@ -0,0 +1,165 @@
...
+     * @param {string | undefined} [expectedFallback] The value that the test
+     * expects the rest parameter of the data client to be equal to.
+     * @param {mocha.Done} [done] The done function used for indicating
+     * that the test is complete or that there is an error in the mocha test
+     * environment.
+     *
+     */
+    function compareRequest(
+      request: DatastoreRequest,
+      expectedFallback: string | undefined,
```
cindy-peng, line 99: Maybe here change the type to Fallback to be consistent too?

cindy-peng (approved):

kolea2 (approved):

murgatroid99 (approved):

looker-colmeia:
hey @danieljbruce thanks for this feature, I would like to ask you a couple of things:

1. you have any estimate of how much percent of query performance in datastore we will loose by enabling `fallback` to `rest`?
2.  this feature will use 100% of requests to datastore using `rest` or it's really some kind of fallback in case of error?

PR #1206, opened 2023-11-30
refactor: Replace common patterns from various transaction methods with common functions

danieljbruce:
**Summary**

This PR reduces code required in the transaction's get, runQuery, runAggregationQuery and commit methods by replacing similar blocks of code with common functions. Reducing the amount of code in each function reduces technical debt and makes it easier to spot decorator patterns that might simplify code even further.

**Changes**

- Added a function called **#sendUserCallbackData** to increase code reuse. This replaces the custom code contained in each call to withBeginTransaction in get, commit, runQuery and runAggregationQuery. Response data from the resolvers in each of these functions is now provided in an array format so that it can be processed by sendUserCallbackData. This lets us eliminate each of the custom callbacks in calls to withBeginTransaction.
- Added a function called **callbackWithError** which eliminates extra code used to build each resolver. This reduces the amount of code required in get, commit, runQuery and runAggregationQuery used to build a resolver because now a callback in the expected format is provided by this function.
- Finally the prior two changes enforced stronger type alignment so in UserCallbackData the error type was changed so that it couldn't be undefined to solve come compiler errors. This is expected because callbacks either receive errors or null values in the first callback argument throughout the codebase so this provides an improvement to how types line up.

### DIFF
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -1040,7 +1040,7 @@ class Transaction extends DatastoreRequest {
    * user that are used for the beginTransaction grpc call.
    * @param {Resolver<T>} [resolver] A resolver object used to construct a
    * custom promise which is run after ensuring a beginTransaction call is made.
-   * @param {(...args: [Error | null, ...T] | [Error | null]) => void} [callback]
+   * @param {Function} [callback]
    * A callback provided by the user that expects an error in the first
    * argument and a custom data type for the rest of the arguments
    * @private
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -233,39 +233,6 @@ class Transaction extends DatastoreRequest {
   }
 
   /**
-   * If the transaction has not begun yet then this function ensures the transaction
-   * has started before running the resolver provided. The resolver is a function with one
-   * argument. This argument is a function that is used to pass errors and
-   * response data back to the caller of the withBeginTransaction function.
-   *
-   * @param {CallOptions | undefined} [gaxOptions] Gax options provided by the
-   * user that are used for the beginTransaction grpc call.
-   * @param {Resolver<T>} [resolver] A resolver object used to construct a
-   * custom promise which is run after ensuring a beginTransaction call is made.
-   * @returns {Promise<UserCallbackData<T>>} Returns a promise that will run
-   * this code and resolve to an error or resolve with the data from the resolver.
-   * @private
-   */
-  async #withBeginTransaction<T>(
-    gaxOptions: CallOptions | undefined,
-    resolver: Resolver<T>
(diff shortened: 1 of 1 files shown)


daniel-sanche:
I'm a little concerned by how complex this change is getting, and this PR seems to be a step in the wrong direction. There are a lot of concepts being used, and they all interact in ways that's hard to follow.

I tried to trace the stack for a commit, and found
`transaction.commit -> sendUserCallback -> withBeginTransaction -> resolver -> runCommit -> callbackWithError -> PromiseResolveFunction -> ...`

From there, it wasn't completely clear to me who was calling the PromiseResolveFunction and where the arguments came from. There's just too many layers of abstraction and callbacks calling other callbacks. There must be a way to simplify this, right?

 If not, maybe adding a lot more comments would be helpful

daniel-sanche:
Some of that feedback applies to PR 1205 as well, but for this one specifically: the change is pretty much line-of-code neutral, and adds two more layers of indirection with `sendUserCallbackData` and `callbackWithError`. So IMO it's probably not worth making this change, unless there are other benefits I'm missing

danieljbruce:
I think this PR cleans up a little bit of technical debt. With the way it was before we are applying the same pattern in four places, but writing the pattern out explicitly each time which we shouldn't do. I see how the indirection could be confusing though.

Overall it isn't too critical to get this in because its just an internal change that cleans things up so if you want to leave it for now then perhaps it is fine.

danieljbruce:
Regarding the stack trace, without callbackWithError you will get a more anonymous callback which might be even harder to debug. We could also think about combining sendUserCallback and withBeginTransaction into one function that has both code and then it would be easier to follow.

src/transaction.ts
```diff
@@ -81,6 +80,31 @@ enum TransactionState {
   IN_PROGRESS, // IN_PROGRESS currently tracks the expired state as well
 }
 
+/**
+ * This function helps build resolvers from common callback data. Callback data
+ * provided by many functions contains error information in the first argument
+ * and an arbitrary response across all other arguments. It is common to build a
```
daniel-sanche, line 86: Would it be possible to make resolver into an object, and replace this with a constructor?

daniel-sanche:
Yeah, IMO this is a step towards more complexity, so probably better to leave it out. I'd look into combining sendUserCallback with withBeginTransaction, and callbackWithError with Resolver. See if you can flatten the call stack. I really don't think it should need this level of complexity

But maybe a node expert would have different thoughts too, so maybe a different reviewer would have different thoughts?

danieljbruce:
Closing this PR in favour of https://github.com/googleapis/nodejs-datastore/pull/1212.

PR #1212, opened 2023-12-11
refactor: remove resolvers, remove promise in withBeginTransaction, simplify code calling withBeginTransaction

danieljbruce:
**Changes:**

Modifies withBeginTransaction so that it doesn't await a promise anymore. Instead, it just calls the function that runs the code we want to run after ensuring the transaction has been started. Since the promise resolved to a UserCallbackData type, this type no longer needs to be used in this function because the promise isn't there anymore. Instead of relying on a value returned by withBeginTransaction and parsing that into callback arguments, withBeginTransaction just calls the callback directly which eliminates another layer in the call stack.

Code calling withBeginTransaction also doesn't use resolvers anymore and logic parsing the return value of withBeginTransaction isn't needed anymore either. This dramatically simplifies code necessary for all functions calling withBeginTransaction.

### DIFF
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -1040,7 +1040,7 @@ class Transaction extends DatastoreRequest {
    * user that are used for the beginTransaction grpc call.
    * @param {Resolver<T>} [resolver] A resolver object used to construct a
    * custom promise which is run after ensuring a beginTransaction call is made.
-   * @param {(...args: [Error | null, ...T] | [Error | null]) => void} [callback]
+   * @param {Function} [callback]
    * A callback provided by the user that expects an error in the first
    * argument and a custom data type for the rest of the arguments
    * @private
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -233,39 +233,6 @@ class Transaction extends DatastoreRequest {
   }
 
   /**
-   * If the transaction has not begun yet then this function ensures the transaction
-   * has started before running the resolver provided. The resolver is a function with one
-   * argument. This argument is a function that is used to pass errors and
-   * response data back to the caller of the withBeginTransaction function.
-   *
-   * @param {CallOptions | undefined} [gaxOptions] Gax options provided by the
-   * user that are used for the beginTransaction grpc call.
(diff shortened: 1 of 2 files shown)


src/transaction.ts
```diff
@@ -1084,6 +1020,41 @@ class Transaction extends DatastoreRequest {
...
+      if (this.#state === TransactionState.NOT_STARTED) {
+        try {
+          await this.#mutex.runExclusive(async () => {
+            if (this.#state === TransactionState.NOT_STARTED) {
+              const runResults = await this.#runAsync({gaxOptions});
+              this.#parseRunSuccess(runResults);
+            }
+          });
+        } catch (err: any) {
+          return callback(err);
```
daniel-sanche, line 1052: What kind of errors might you encounter here? A comment would be useful

src/transaction.ts
```diff
@@ -1084,6 +1020,41 @@ class Transaction extends DatastoreRequest {
...
+    gaxOptions: CallOptions | undefined,
+    fn: () => void,
+    callback: (...args: [Error | null, ...T] | [Error | null]) => void
+  ): void {
+    (async () => {
+      if (this.#state === TransactionState.NOT_STARTED) {
+        try {
+          await this.#mutex.runExclusive(async () => {
+            if (this.#state === TransactionState.NOT_STARTED) {
+              const runResults = await this.#runAsync({gaxOptions});
```
daniel-sanche, line 1047: nit: it looks like runAsync calls `beginTransaction`. Maybe name it something like `beginTransactionAsync`?

daniel-sanche (approved):
This is a great change! Just a couple comments

src/transaction.ts
danieljbruce: Added.

src/transaction.ts
danieljbruce: Sure, changed to beginTransactionAsync. Since this is in the transaction class run typically means run the transaction, but maybe beginTransactionAsync is more descriptive. Either way, this method is private so the name can be modified later if need be.

PR #1229, opened 2024-01-26
test: remove await from code samples

danieljbruce:
**General changes:**

A customer raised an issue where they ran the OR query sample from https://cloud.google.com/datastore/docs/samples/datastore-query-filter-or and an error was thrown on the line that starts with `await`. `await` can only be used inside an async function so the code sample should not contain `await` when it is not in an async function. This PR removes await from code samples where await is not used inside an async function.

**OR query sample changes:**

In the OR query sample tests, the assert check no longer works after removing await as console.log.firstCall now equals null. This is because console log output inside an async function is not visible outside of the function if await is not used. Therefore, we should use the execSync function like the code in the import-export.test.js sample test file to get console output from running the sample. This way the `it` block with the assert statement can read the console output and ensure it equals “Entity found: Feed cats\nEntity found: Buy milk\n”.

It is also possible that the database currently contains “Buy milk” and “Feed cats” data before the code sample is run which will affect the console output when running the OR query sample. Therefore, a script is added which runs before the OR query sample test is run. This script removes “Buy milk” and “Feed cats” data from the database so that the database starts from a consistent state and the console output is consistent in the assertion check.

### DIFF
--- a/samples/test/queryFilterOr.test.js
+++ b/samples/test/queryFilterOr.test.js
@@ -23,8 +23,10 @@ const sinon = require('sinon');
 const {Datastore} = require('@google-cloud/datastore');
 const datastore = new Datastore();
 
-const {queryFilterOr} = require('../queryFilterOr');
+const cp = require('child_process');
 let taskKey1, taskKey2;
+
+const execSync = cmd => cp.execSync(cmd, {encoding: 'utf-8'});
 
 describe('Creating a union query', () => {
   const stubConsole = function () {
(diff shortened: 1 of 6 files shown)


bhshkh (approved):

samples/test/queryFilterOr.test.js
```diff
@@ -67,7 +81,7 @@ describe('Creating a union query', () => {
   });
 
   it('should get a combination of items from the Datastore', async () => {
-    await queryFilterOr();
-    assert.include(console.log.firstCall.args[0], 'Entity');
+    const stdout = execSync(`node ./queryFilterOr.js`);
```
cindy-peng, line 84: Nit: Inline simple local variables: go/unit-testing-practices#example-inlining-simple-local-variables

cindy-peng (approved):

PR #1239, opened 2024-03-04
feat: new transaction feature

danieljbruce:
**Summary**

This PR ensures that transaction read calls will use [newTransaction options](https://github.com/googleapis/nodejs-datastore/blob/1585d4a4e1b4b16d198307a3e97ffcf156d000b1/protos/protos.d.ts#L7588) to begin transactions instead of making a separate `beginTransaction` grpc call.

**Source code changes**

A `#blockWithMutex` function is added and it is used for read calls instead of a `#withBeginTransaction` function. This makes it so that all read calls on Transaction will block on the mutex and then call the same method in their super class. Before this change, the read calls on Transaction would block on the mutex, make a `beginTransaction` grpc call and then call the same method in their super class. So this change to using `#blockWithMutex` omits the extra `beginTransaction` call and instead expects the function it is calling to supply new transaction options.

A function `getTransactionRequest` is added which pulls out the code used for building transaction options for a `beginTransaction` request. That is because this code is also used to build the new transaction options.

TransactionState is moved up to the super class `DatastoreRequest` and a new state is added called `NOT_TRANSACTION` for `DataRequests` that are not transactions.

`parseTransactionResponse` is moved up to the super class `DatastoreRequest` and is now used by `runQuery`, `runAggregationQuery` and `get` to save the transaction id.

`newTransaction` options are now built in `getRequestOptions` which is used by `runQuery`, `runAggregationQuery` and `get` to save the transaction id.

`state` is moved up to the `DatastoreRequest` super class because it now has to be used to decide whether or not to build newTransaction options.

If a transaction is expired then an error will be thrown when the transaction is used.

ReadOptions are the options shared by Datastore read requests.
[new_transaction](https://github.com/googleapis/go-genproto/blob/478b75c54725e54e5f92966212c6572c091596aa/googleapis/datastore/v1/datastore.pb.go#L1794-L1803) is a [consistency type](https://github.com/googleapis/go-genproto/blob/478b75c54725e54e5f92966212c6572c091596aa/googleapis/datastore/v1/datastore.pb.go#L1708) in ReadOptions. When new_transaction is used in a read request, a new transaction is begun for the request.

**System tests added**

A system test is added to ensure using new transaction has better latency. System tests are added for runQuery, runAggregationQuery and get for readOnly transactions to ensure readOnly transactions work correctly whether `transaction.run` is called or not.

**Unit tests added**

Unit tests are added for various orders of transaction operations that ensure correct behaviour for when `transaction.run` is used and when `transaction.run` is omitted. `lookup, lookup, put, commit` is one order that is tested. The others are `runQuery, lookup, put, commit`, `runAggregationQuery, lookup, put, commit`, `put, put, lookup, commit` and `put, commit`. A unit test is also corrected.

**Drawbacks**

If the user uses `transaction.runQueryStream` or they use `transaction.createReadStream` then the call will not block on a mutex. These methods are not async methods so they cannot wait for a `#blockWithMutex` call to complete. If users are currently using `runQueryStream` or `createReadStream` concurrently then these multiple read calls may be beginning multiple transactions at the same time using one transaction object which is not ideal.


src/request.ts
```diff
@@ -1116,6 +1188,36 @@ class DatastoreRequest {
   }
 }
 
+function isTransaction(request: DatastoreRequest): request is Transaction {
+  return request instanceof Transaction;
+}
+
+export function getTransactionRequest(
```
danieljbruce, line 1195: This function is used by `DatastoreRequest` in `request.ts` and `Transaction` in `transaction.ts`. `Transaction` inherits from `DatastoreRequest` so even though this could be moved to a protected method in `DatastoreRequest`, we want to avoid doing so. This is because adding the `protected` reserved word in Typescript doesn't entirely hide the method from the user in the class like `protected` should.

src/request.ts
```diff
@@ -89,6 +109,7 @@ class DatastoreRequest {
     | Array<(err: Error | null, resp: Entity | null) => void>
     | Entity;
   datastore!: Datastore;
+  protected state: TransactionState = TransactionState.NOT_TRANSACTION;
```
danieljbruce, line 112: I don't like using `protected` here because typescript doesn't technically hide `protected` properties completely, but it seems that there is no simple way around it because both `Transaction` and `DatastoreRequest` need access to the `state` variable. There is more information about some complex ways to avoid this in the design doc in Alternative 3, but they require major architecture changes.

src/request.ts
```diff
@@ -990,16 +1061,34 @@ class DatastoreRequest {
...
+          {}
+        );
+        sharedQueryOpts.readOptions.consistencyType = 'newTransaction';
+      }
+    }
     if (options.consistency) {
       const code = CONSISTENCY_PROTO_CODE[options.consistency.toLowerCase()];
-      sharedQueryOpts.readOptions = {
-        readConsistency: code,
-      };
```
danieljbruce, line 997: This was a bug because it would overwrite `newTransaction` if `readConsistency` was provided.

src/request.ts
```diff
@@ -1122,18 +1211,27 @@ class DatastoreRequest {
...
       (method === 'lookup' ||
         method === 'runQuery' ||
         method === 'runAggregationQuery')
     ) {
-      if (reqOpts.readOptions && reqOpts.readOptions.readConsistency) {
-        throw new Error(
-          'Read consistency cannot be specified in a transaction.'
-        );
-      }
-
```
danieljbruce, line 1136: This is just moved up to also be included for new transaction calls. Also, the error was not propogating up and needs to be wrapped in a callback.

src/transaction.ts
```diff
@@ -53,11 +54,6 @@ interface BeginAsyncResponse {
   resp?: google.datastore.v1.IBeginTransactionResponse;
 }
 
-enum TransactionState {
-  NOT_STARTED,
-  IN_PROGRESS, // IN_PROGRESS currently tracks the expired state as well
-}
-
```
danieljbruce, line 60: Moved to the super class because it is now needed there.

src/transaction.ts
```diff
@@ -696,24 +684,10 @@ class Transaction extends DatastoreRequest {
...
-      };
-    }
-
-    if (options.transactionOptions) {
-      reqOpts.transactionOptions = options.transactionOptions;
-    }
     return new Promise((resolve: (value: BeginAsyncResponse) => void) => {
+      const reqOpts = {
+        transactionOptions: getTransactionRequest(this, options),
+      };
```
danieljbruce, line 690: `getTransactionRequest` does what the removed code does here.

test/gapic-mocks/runQuery.ts
```diff
@@ -68,35 +53,13 @@ describe('Run Query', () => {
     }
   }
 
-  it('should pass read time into runQuery for transactions', async () => {
```
danieljbruce, line 71: Note that this test is no longer valid because transaction and read time are no longer compatible together as is the case in Python.

src/transaction.ts
```diff
@@ -434,16 +430,32 @@ class Transaction extends DatastoreRequest {
...
-    this.request_(
-      {
-        client: 'DatastoreClient',
-        method: 'rollback',
-        gaxOpts: gaxOptions || {},
+    if (this.state === TransactionState.EXPIRED) {
+      callback(new Error(transactionExpiredError));
+      return;
+    }
+    if (this.state === TransactionState.NOT_STARTED) {
```
cindy-peng, line 437: Just wondering what are the allowed states here?  NOT_TRANSACTION and IN_PROGRESS?

src/request.ts
```diff
@@ -846,6 +929,10 @@ class DatastoreRequest {
 
     let info: RunQueryInfo;
 
+    if (this.state === TransactionState.EXPIRED) {
+      callback(new Error(transactionExpiredError));
+      return;
+    }
```
cindy-peng, line 935: Do we need to check `readTimeAndConsistencyError` here?

src/request.ts
```diff
@@ -335,6 +366,15 @@ class DatastoreRequest {
...
+   */
+  protected checkExpired() {
+    if (this.state === TransactionState.EXPIRED) {
+      throw Error(transactionExpiredError);
+    }
+  }
+
   /**
    * Retrieve the entities as a readable object stream.
    *
```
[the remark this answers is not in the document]
danieljbruce: Sure. Done.

src/transaction.ts
danieljbruce: NOT_TRANSACTION, NOT_STARTED, IN_PROGRESS, EXPIRED.

src/request.ts
danieljbruce: No. That check is done further downstream in the call to `runQueryStream`. But I did spot some other cases where errors weren't bubbling up and fixed those bugs.

daniel-sanche (approved):
LGTM

src/request.ts
```diff
@@ -1229,6 +1328,54 @@ class DatastoreRequest {
...
+  if (options.transactionOptions) {
+    reqOpts = {};
+    if (options.transactionOptions.readOnly) {
+      reqOpts.readOnly = {};
+    }
+    const id = options.transactionOptions.id;
+    if (id) {
+      reqOpts.readWrite = {previousTransaction: id};
+    }
+  }
```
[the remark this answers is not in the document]
danieljbruce: I generated unit tests, corrected them and then reduced this function down to one ternary operation. Previously this was just copy pasted code from before. Search `describe('getTransactionRequest', () => {` to see how this function should behave.

src/request.ts
```diff
@@ -1229,6 +1328,54 @@ class DatastoreRequest {
...
+    if (options.readOptions && options.readOptions.readTime) {
+      throw new Error('Read time cannot be specified in a transaction.');
+    }
+  }
+}
+
+export function getTransactionRequest(
+  transaction: Transaction,
+  options: RunOptions
+): TransactionRequestOptions {
```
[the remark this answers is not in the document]
danieljbruce: Done

sofisl (approved):

PR #1253, opened 2024-05-27
docs(samples): Query profiling feature samples

danieljbruce:
**Changes:**

In _samples/queryFilterOr.js_, the sort isn't working properly. This results in a flakey test for OR queries because runQuery doesn't guarantee an order for results and relies the sort so that the results are in the right order for the assert statement. The sort function is fixed so it now works properly.

Samples files _samples/queryProfileExplain.js_, _samples/queryProfileExplainAggregation.js_, _samples/queryProfileExplainAnalyze.js_, _samples/queryProfileExplainAnalyzeAggregation.js_ are added which contain the code samples that execute runQuery and runAggregationQuery for both query planning and query profiling. The filenames match the filenames in java-datastore.

In _samples/test/queryFilterOr.test_ we pull out the code that sets the database to a specific state into a _samples/test/helpers/populate-data.js_ file because we want to reuse this code to set the database to a specific state for the query profiling samples as well.

In _samples/test/queryProfiling.test.js_ we add samples tests that run each of the 4 files with query profiling code samples and compare the output in each sample to the expected output for each code sample.


samples/queryProfileExplain.js
```diff
@@ -0,0 +1,38 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+// sample-metadata:
+//   title: Run query profiling (regular query)
+//   description: Run query profiling for a standard query and print query results
```
kolea2, line 19:
```suggestion
//   description: Run query explain for a standard query and print query explain metrics
```

kolea2: (here and in other places, please update)

samples/queryProfileExplain.js
```diff
@@ -0,0 +1,38 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+// sample-metadata:
+//   title: Run query profiling (regular query)
```
kolea2, line 18:
```suggestion
//   title: Run query explain (regular query)
```

kolea2: please update elsewhere, too

samples/queryProfileExplainAnalyze.js
```diff
@@ -0,0 +1,44 @@
...
+  // Imports the Cloud Datastore
+  const {Datastore} = require('@google-cloud/datastore');
+
+  // Instantiate the Datastore
+  const datastore = new Datastore();
+  const q = datastore.createQuery('Task');
+  const [entities, info] = await datastore.runQuery(q, {
+    explainOptions: {analyze: true},
+  });
+  entities.sort((e1, e2) => {
```
kolea2, line 33: why is sorting necessary?

danieljbruce: Much like the [or query test](https://github.com/googleapis/nodejs-datastore/blob/b7ff5c86306f80d93f678a0638892c58a3b2088c/samples/test/queryFilterOr.test.js#L84) the entities are expected to be in a specific order for the test output. `runQuery` doesn't guarantee entities in a certain order so if we don't sort them then we end up with a flakey test.

kolea2: if this is only meant to assist in tests, I would prefer we change the test itself to assert on the contents of the results, ignoring the order.

samples/queryProfileExplainAggregation.js
```diff
@@ -0,0 +1,42 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+// sample-metadata:
+//   title: Run query profiling (aggregation query)
```
kolea2, line 18: please fix to match other title (and in other places)

kolea2 (dismissed):
looking good! A few small TODOs.

samples/queryProfileExplainAnalyze.js
danieljbruce: Done

samples/queryProfileExplainAggregation.js
danieljbruce: Done

samples/test/helpers/populate-data.js
```diff
@@ -0,0 +1,56 @@
...
+  taskKey2 = datastore.key('Task');
+  const entity2 = {
+    key: taskKey2,
+    data: {
+      description: 'Feed cats',
+    },
+  };
+
+  // Ensure the datastore database has no existing data.
+  const query = datastore.createQuery('Task');
```
cindy-peng, line 45: Nit: does it make more sense to add these clean up to after() in L53?

cindy-peng (approved):

samples/queryProfileExplain.js
```diff
@@ -0,0 +1,38 @@
...
+  // Imports the Cloud Datastore
+  const {Datastore} = require('@google-cloud/datastore');
+
+  // Instantiate the Datastore
+  const datastore = new Datastore();
+  const q = datastore.createQuery('Task');
+  const [, info] = await datastore.runQuery(q, {
+    explainOptions: {analyze: false},
+  });
+  console.log(`info: ${Object.keys(info.explainMetrics)}`);
```
bhshkh, line 33: When I run this code, I'm seeing error: 
```
client.ts:32
      console.log(`info: ${Object.keys(info.explainMetrics)}`);
                                  ^

TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at quickstart (/usr/local/google/home/bahaaiman/Documents/cfdb-workspace-01/clients/node/nodejs-datastore-client/client.ts:32:35)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)

Node.js v20.14.0
```

What am I doing wrong in the code?
```
// Imports the Google Cloud client library
const { Datastore } = require('@google-cloud/datastore');

// Creates a client
const datastore = new Datastore({projectId: "my-project"});

async function quickstart() {
    // The Cloud Datastore key for the new entity
    const taskKey = datastore.key(['Task', 'task1']);

    // Prepares the new entity
    const task = {
        key: taskKey,
        data: {
            Description: 'Buy milk2',
            Done: false,
            Slice: [100, 200, 300]
        },
    };

    // Saves the entity
    await datastore.save(task);
    console.log(`Saved ${task.key.name}: ${task.data.Description}`);

    const query = datastore.createQuery('Task');
    const [entities, info] = await datastore.runQuery(query, {
        explainOptions: {analyze: true},
      });
      for (const entity of entities) {
        console.log(`Entity found: ${entity['description']}`);
      }
      console.log(`info: ${Object.keys(info.explainMetrics)}`);

}
quickstart();
```

samples/test/helpers/populate-data.js
danieljbruce: No because then the samples tests won't work. This code is responsible for clearing the database and setting it to a specific state so that when the samples run the output is a predictable value.

samples/queryProfileExplain.js
bhshkh: Checked with @danieljbruce offline. Updating the datastore version to 9.0.0 gets rid of the error

bhshkh (approved):

Issue #1242, opened 2024-03-21
Unable to exclude large indexes


dogmatic69:
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

1) Is this a client library issue or a product issue?
client lib

2) Did someone already solve this?
existing feature that is broken

3) Do you have a support contract?
no

#### Environment details

  - OS: Alpine
  - Node.js version: 21.7.1
  - yarn: 1.22.19
  - `@google-cloud/datastore` version: 8.5.0

#### Steps to reproduce

in the [code](https://github.com/googleapis/nodejs-datastore/blob/main/src/entity.ts#L993-L995), when passing data with the name/value format rather than a normal object, this functionality seems to be completely skipped. 

Simple save function with the `excludeLargeProperties` prop hard coded to true results in error that a field is too large. Given the prop is true, I'd expect it to ignore any large fields.

Feasible to include this functionality or do I avoid using the verbose data structure when saving? 

```
await instance.save({
  key,
  data: [
    {
      name: 'field_a',
      value: 'value_a',
    },
    {
      name: 'field_b',
      value: {
        nested_field: 'nested_value', // completely ignored
      },
    }
  ],
  excludeLargeProperties: true,
})
```

The above code with a suitably large field results in errors like this
```
Error saving data: 3 INVALID_ARGUMENT: The value of property "content" is longer than 1500 bytes. Error: 3 INVALID_ARGUMENT: The value of property "content" is longer than 1500 bytes.
    at callErrorFromStatus (/app/node_modules/@grpc/grpc-js/build/src/call.js:31:19)
    at Object.onReceiveStatus (/app/node_modules/@grpc/grpc-js/build/src/client.js:192:76)
    at Object.onReceiveStatus (/app/node_modules/@grpc/grpc-js/build/src/client-interceptors.js:360:141)
    at Object.onReceiveStatus (/app/node_modules/@grpc/grpc-js/build/src/client-interceptors.js:323:181)
    at /app/node_modules/@grpc/grpc-js/build/src/resolving-call.js:99:78
    at process.processTicksAndRejections (node:internal/process/task_queues:77:11)
for call at
    at ServiceClientImpl.makeUnaryRequest (/app/node_modules/@grpc/grpc-js/build/src/client.js:160:32)
[...]

PR #1265, opened 2024-08-19
fix: Handle excludeLarge properties for arrays properly

danieljbruce:
**Summary**

This PR contains the fix for solving [this issue](https://github.com/googleapis/nodejs-datastore/issues/1242) where `excludeLargeProperties: true` doesn't work properly for arrays. It also ensures that the fix doesn't break existing use cases.

**Background**

This PR is a successor to [this other PR](https://github.com/googleapis/nodejs-datastore/pull/1263) and contains a fix that addresses the failing test cases in the other PR.

**Changes**

- `entityToEntityProto` used to have functionality that would add `excludeFromIndexes: true` in the right places, but we move this code into a separate `addExcludeFromIndexes` function because we want to reuse it for arrays.
- In `save` the code that calls `findLargeProperties_` is moved so that it is only used as is for non array data. For the array data this `findLargeProperties_` code is used per element to populate the `excludeFromIndexes` list in the correct manner.
- For array data, now that `excludeFromIndexes` list is correctly calculated we can use the `addExcludeFromIndexes` functionality that we pulled out to apply `excludeFromIndexes: true` in the right places.

**Next Steps**

- Add an integration test containing the code mentioned in the issue.
- Add integration tests that uses the JSON array of tests from this PR.
- Make the save function more modular.
- Add some more test cases just to be sure

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -69,6 +69,7 @@ import {google} from '../protos/protos';
 import {AggregateQuery} from './aggregate';
 
 const {grpc} = new GrpcClient();
+const addExcludeFromIndexes = entity.addExcludeFromIndexes;
 
 export type PathType = string | number | entity.Int;
 export interface BooleanObject {
@@ -1130,12 +1131,6 @@ class Datastore extends DatastoreRequest {
               acc: EntityProtoReduceAccumulator,
               data: EntityProtoReduceData
             ) => {
-              /*
-              const subEntityObject = {
-                data:
-                excludeFromIndexes: [],
-              };
-               */
               const value = entity.encodeValue(
                 data.value,
                 data.name.toString()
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -1125,6 +1118,25 @@ class Datastore extends DatastoreRequest {
         // This was replaced with a more efficient mechanism in the top-level
         // `excludeFromIndexes` option.
```
daniel-sanche, line 1119: is this TODO still relevant? Does the top-level excludeFromIndexes work the same way as these changes?

daniel-sanche (approved):

src/index.ts
danieljbruce: I think this TODO can be removed. There is currently code [here](https://github.com/googleapis/nodejs-datastore/blob/bbcfc3e73c56e3aa8152ecc9395d3c94133ea6b4/src/index.ts#L1138-L1150) that adds excludeFromIndexes in the top level and I added code that provides `excludeFromIndexes` everywhere else based on the path that gets generated. To avoid test failures, the code for adding `excludeFromIndexes` in the top level is still needed so I don't think this TODO is relevant anymore.

Issue #1327, opened 2024-10-01
Should not throw TypeError when saving `null` for a field with subproperty index-exclusion


mrnagydavid:
#### Environment details

  - OS: N/A
  - Node.js version: N/A
  - npm version: N/A
  - `@google-cloud/datastore` version: v9.1.0, but in earlier versions too

#### Problem statement

```js
save({ excludeFromIndexes: ['foo.bar'], entity: { sho: 1, foo: null })
// throws TypeError: Cannot read properties of undefined (reading 'properties')

save({ excludeFromIndexes: ['foo[].*'], entity: { sho: 1, foo: null })
// throws TypeError: Cannot read properties of undefined (reading 'values')
```

Considering the following 4 types of index exclusions:
1. `excludeFromindexes: ['foo.*']`
1. `excludeFromindexes: ['foo.bar']`
1. `excludeFromindexes: ['foo[].*']`
1. `excludeFromindexes: ['foo[].bar']`

Supplying `{ entity: { foo: null } }` throws for 2 and 3, but not for 1 and 4.

It should either throw in all cases or in none. I think it should throw in none, and it is easily fixable.

#### Steps to reproduce

Easily reproducable via unit tests:
[repo with unit test showcasing error](https://github.com/googleapis/nodejs-datastore/compare/main...mrnagydavid:nodejs-datastore:index-exclusion-problem-statement?expand=1#diff-f65d8e4f4c11798cbfa93ffb45060359589abc6098920c09a210c3fa17e70529)

#### Proposed solution

```ts
// entity.ts

function excludePathFromEntity(entity: EntityProto, path: string) {
  if (!entity) return; // <-- !! fixes 'foo.bar'
  // ...
  } else if (
    firstPathPartIsArray &&
    hasWildCard &&
    remainderPath === '*' &&
    isFirstPathPartDefined
  ) {
    const array = entity.properties![firstPathPart].arrayValue;
    if (!array) return; // <-- !! fixes 'foo[].*'
  // ...
```

This TypeError can be fixed with 2 lines of code and I am ready to [open a PR](https://github.com/googleapis/nodejs-datastore/compare/main...mrnagydavid:nodejs-datastore:fix/fix-saving-null-value-for-index-excluded-fields?expand=1).

danieljbruce:
@mrnagydavid 
Instead of modifying the test, please add a new test that matches what you changed the old test to. After that it LGTM.

mrnagydavid:
[...]

PR #1330, opened 2024-10-04
fix: not check index exclusion path for `null` values

mrnagydavid:
This PR fixes issue #1327  🦕

### Summary

Considering the following 4 types of index exclusions:
1. `excludeFromindexes: ['foo.*']`
1. `excludeFromindexes: ['foo.bar']`
1. `excludeFromindexes: ['foo[].*']`
1. `excludeFromindexes: ['foo[].bar']`

Supplying `{ entity: { foo: null } }` throws for 2 and 3, but not for 1 and 4.

```js
save({ excludeFromIndexes: ['foo.bar'], entity: { sho: 1, foo: null })
// throws TypeError: Cannot read properties of undefined (reading 'properties')

save({ excludeFromIndexes: ['foo[].*'], entity: { sho: 1, foo: null })
// throws TypeError: Cannot read properties of undefined (reading 'values')
```

With the changes in this PR, the save calls will not throw in any of the 4 cases.


danieljbruce (approved):

danieljbruce:
For additional notes on this PR see [this](https://github.com/googleapis/nodejs-datastore/pull/1361).

PR #1336, opened 2024-10-23
fix: The save function should not throw an error for name/value properties in arrays

danieljbruce:
**Summary:**

When calling `datastore.save` an error will be thrown for input arguments that have nested arrays with objects containing `name` and `value` properties when the excludeLargeProperties flag is set to true. Such input arguments are valid input arguments for the save function and the right properties should be excluded from indexes so that an error doesn't occur. This PR is also just onto a feature branch so after `Next Steps` are completed for this feature branch there will be an opportunity to review everything together.

**Background:**

Previously [a fix](https://github.com/googleapis/nodejs-datastore/pull/1266) was applied that would allow calling save to not throw an error when providing large values for nested fields and setting `excludeLargeProperties` to true. However, unique edge cases remain where `save` throws an error for requests that should be ok and `excludeFromIndexes: true` gets applied in the wrong places.

When a user calls `save`, the following steps are followed and this PR changes the `findLargeProperties_` function used in step 4 below:
1. The user passes `entities` into the save function
2. For each entity the steps 3-6 happen:
3. Some preprocessing is done
4. If `excludeLargeProperties: true` is provided then the excludeFromIndexes list is extended using the `findLargeProperties_` function. This step uses the `findLargeProperties_` function differently depending on whether the entity is an array or a non-array.
5. An entityProto is built from the entity. This is done differently depending on whether the entity is an array or the entity is a non-array.
6. A mutation is built from the entityProto
7. All mutations are sent to the server

**Changes:**

- For **`system-test/datastore.ts`** an integration test is added that calls `save` with long `name/value` properties inside a nested array. Without the source code changes from this PR, this test fails because the `save` call throws an error. This test ensures that the save call doesn't throw an error. A second test is also added just to be sure that long strings will be excluded from indexes when they are in the top level of the array.
- For **`test/gapic-mocks/commit.ts`** a unit test is added that makes the same `save` call as the new integration test in **`system-test/datastore.ts`**. It ensures that `excludeFromIndexes: true` is added in all the right places for the mutations proto.
- For **`test/entity/findLargeProperties_.ts`** we write some unit tests for the `findLargeProperties_` function because that function is where the source code changes are in this PR. Also, there currently are no unit tests for the `findLargeProperties_` function and the function is complex so the unit tests are helpful for documenting what the function actually does.
- For **src/entity.ts** some code is removed that pertains to objects with name/value properties in arrays. The removed code adds `excludeFromIndexes: true` to various locations in the `entityObject` which isn't necessary anymore on the top level because that is done [in the save function](https://github.com/googleapis/nodejs-datastore/blob/1d5643369226c5fc220779f4c90fa804d8f876af/src/index.ts#L1120-L1136). It also isn't something we want to do in levels other than the top level because the `name/value` encoding for array entity protos is in the top level. This code is also preventing a recursive search for large values in child objects when an object in an array happens to have `name` and `value` properties, but we want this recursive search for large values so the code should be removed.

**Next Steps:**

There are so many different combinations of input for the save function that it can be difficult to have complete test coverage. I suggest we break the save function down into smaller parts?

Right now two of the steps in the save function are:
4. If `excludeLargeProperties: true` is provided then the excludeFromIndexes list is extended using the `findLargeProperties_` function. This step uses the `findLargeProperties_` function differently depending on whether entities are arrays or non-arrays.
5. An entityProto is built from an entity. This is done differently depending on whether the entity is an array or the entity is a non-array.

We should do the following on `exclude-from-indexes-cleanup` branch:
1. Extract step 4 out into a function so that it can be unit tested.
2. Extract step 5 out into a function so that it can be unit tested.
3. Ensure that every input to step 4 and then step 5 results in an entity proto where `excludeFromIndexes: true` is only in all places where a long string is present.


src/entity.ts
```diff
@@ -991,16 +991,6 @@ export namespace entity {
...
     if (Array.isArray(entities)) {
       for (const entry of entities) {
-        if (entry && entry.name && entry.value) {
-          if (
-            is.string(entry.value) &&
-            Buffer.from(entry.value).length > MAX_DATASTORE_VALUE_LENGTH
-          ) {
-            entry.excludeFromIndexes = true;
-          } else {
-            continue;
```
danieljbruce, line 1001: We don't want this `continue` because then the `findLargeProperties_` method call below won't happen and we won't search nested properties of `entry`  for large properties and add them to the list of properties to exclude.

src/entity.ts
```diff
@@ -991,16 +991,6 @@ export namespace entity {
     const MAX_DATASTORE_VALUE_LENGTH = 1500;
     if (Array.isArray(entities)) {
       for (const entry of entities) {
-        if (entry && entry.name && entry.value) {
```
danieljbruce, line 994: The original intent of this `if` statement was to deal with arrays passed into the save function. When arrays are passed into the save function, they are normally expected to have `name` and `value` properties defined at the top level and then [their proto is encoded using the name as the key](https://github.com/googleapis/nodejs-datastore/blob/519e33b3c69c64cfcd4a349c507bf9de85697fdb/src/index.ts#L1143-L1164) so this code adds `excludeFromIndexes:true` in the right places for this special encoding.

The problem with this if statement is that this function is recursive so this logic will be applied recursively, but we only want this logic to be applied at the top level for arrays because the special encoding is only done for arrays at the top level.

src/entity.ts
```diff
@@ -991,16 +991,6 @@ export namespace entity {
...
-          if (
-            is.string(entry.value) &&
-            Buffer.from(entry.value).length > MAX_DATASTORE_VALUE_LENGTH
-          ) {
-            entry.excludeFromIndexes = true;
```
danieljbruce, line 999: Leaving this assignment in will apply `excludeFromIndexes: true` in more places than is necessary. As mentioned in the comment above, we already apply `excludeFromIndexes` in the right places elsewhere in the top level for arrays and in lower levels this code will actually apply `excludeFromIndexes: true` to the wrong places so it should be removed.

system-test/datastore.ts
```diff
@@ -348,6 +348,37 @@ async.each(
...
+                {
+                  name: longString,
+                  value: 'some-value',
+                },
+              ],
+            },
+            excludeLargeProperties: true,
+          });
+        });
+        it('should save a nested name/value pair with value as a long string', async () => {
```
danieljbruce, line 367: This test shows that even though we are omitting the code that appends `entry.excludeFromIndexes = true;` to the entity that `entry.excludeFromIndexes = true;` is still applied in the right places for entities that are arrays.

daniel-sanche (approved):
LGTM

PR #1346, opened 2024-10-29
fix(doc): Fix param descriptions for filters

cindy-peng:
Fix issue the returns value doesn't contain a description. The fix is to add a description.

Link to Example: https://cloud.google.com/nodejs/docs/reference/datastore/latest/overview#_google_cloud_datastore_and_function_1_


src/filter.ts
```diff
@@ -34,18 +34,26 @@ enum CompositeOperator {
 }
 
 /**
- * Returns an AND composite filter.
+ * And filters are composed of many other filters and when they are applied
+ * then query results are only returned when they pass through all these other
+ * filters.
  *
- * @param {EntityFilter[]} filters The filters that make up the AND filter.
+ * @param {EntityFilter[]} filters The list of filters that make up the and filter.
```
danieljbruce, line 41: Consider making AND capital here.

src/filter.ts
```diff
@@ -34,18 +34,26 @@ enum CompositeOperator {
...
 }
 
 /**
- * Returns an OR composite filter.
+ * Or filters are composed of many other filters and when they are applied
+ * then query results are returned when they pass through any of these other
+ * filters.
  *
- * @param {EntityFilter[]} filters The filters that make up the OR filter.
+ * @param {EntityFilter[]} filters The list of filters that make up the or filter.
```
danieljbruce, line 54: Consider making OR capital here.

danieljbruce (approved):
Just run the linter and then this should be good to go.

cindy-peng:
> linter

Hi Dan, do you know how can I run the linter?

PR #1349, opened 2024-10-31
refactor: Add separate function for adding large properties to the excludeFromIndexes list

danieljbruce:
**Summary:**

Currently in the save method, if the user sets excludeLargeProperties to true then some code in this method will extend the excludeFromIndexes list based on the position of large properties in the entity object. This PR refactors all of this code into a separate function called `extendExcludeFromIndexes` because it allows us to unit test this code to make sure it is working correctly. We cannot unit test this code when it is buried inside the save function like it is right now and it allows us to complete the next steps listed below.

This is a follow-up to [this PR](https://github.com/googleapis/nodejs-datastore/pull/1336).

**Next steps:**

Next we need to do the following:
1. Pull the code that creates an entity proto from an entity object currently inside the save function into a separate function so that it can be unit tested. Call this new internal function `buildEntityProto`.
2. Write tests to ensure that for any input, whenever `extendExcludeFromIndexes` is used and then `buildEntityProto` is used, the resulting entity proto always has `excludeFromIndexes: true` in all the places of the entity proto where the large properties are.


src/utils/entity/extendExcludeFromIndexes.ts
```diff
@@ -0,0 +1,52 @@
...
+// limitations under the License.
+
+import {entity, Entity} from '../../entity';
+
+/**
+ * This function extends the excludeFromIndexes list when it finds
+ * large properties in the entity object. The extended excludeFromIndexes
+ * list is then used when building the entity proto.
+ */
+// TODO: Add params
```
daniel-sanche, line 22: is this todo still valid?

src/utils/entity/extendExcludeFromIndexes.ts
```diff
@@ -0,0 +1,52 @@
...
+export function extendExcludeFromIndexes(entityObject: Entity) {
+  if (entityObject.excludeLargeProperties) {
+    if (Array.isArray(entityObject.data)) {
+      // This code populates the excludeFromIndexes list with the right values.
+      if (entityObject.excludeLargeProperties) {
```
daniel-sanche, line 27: isn't this redundant from line 24?

src/utils/entity/extendExcludeFromIndexes.ts
danieljbruce: I just pushed updates to address this. This is about adding documentation for the parameters which is done now.

src/utils/entity/extendExcludeFromIndexes.ts
danieljbruce: Yup. I just pushed an update to remove this line.

daniel-sanche (approved):
LGTM

PR #1350, opened 2024-10-31
refactor: Add a separate function for building the entity proto

danieljbruce:
**Summary:**

Currently in the save method, the entity objects that the user passed in will be converted to entity protos. The save method has [code that does this for arrays](https://github.com/googleapis/nodejs-datastore/blob/536873e24bacc9477f1a9c4c5403ed08d5c8cc93/src/index.ts#L1138-L1166) and [code that does this for non-arrays](https://github.com/googleapis/nodejs-datastore/blob/536873e24bacc9477f1a9c4c5403ed08d5c8cc93/src/index.ts#L1180). This PR refactors all of this code into a separate function called `buildEntityProto` because it allows us to unit test this code to make sure it is working correctly. We cannot unit test this code when it is buried inside the save function like it is right now and it allows us to complete the next steps listed below.

This is a follow-up to [this PR](https://github.com/googleapis/nodejs-datastore/pull/1349).

**Changes**

- In `src/index.ts`, code inside the save function is moved out into `buildEntityProto.ts` so that it can be unit tested.
- In `src/utils/entity/buildEntityProto.ts` the `buildEntityProto` function is created that now contains the code that used to be in `src/index.ts`.
- In `test/entity.ts`, some constants in a test are moved out into a `fixtures` file so that they can be reused by another test.
- In `test/entity/buildEntityProto.ts`, some tests have been created for the `buildEntityProto` function and some of those tests use the constants that have been moved to the `fixtures` file.
- In `test/fixtures/entityObjectAndProto.ts`, constants are added that used to be in `test/entity.ts`.
- In `test/index.ts`, new mocks need to be added to the proxyquire object due to the fact that the location of the code that builds the entity proto is now in `src/utils/entity/buildEntityProto.ts`. So mocks that used to apply for `src/index.ts` will now apply for `../src/utils/entity/buildEntityProto.js`.

**Next steps:**

Next we need to write tests to ensure that for any input, whenever `extendExcludeFromIndexes` is used and then `buildEntityProto` is used, the resulting entity proto always has excludeFromIndexes: true in all the places of the entity proto where the large properties are.


src/index.ts
```diff
@@ -1117,44 +1116,7 @@ class Datastore extends DatastoreRequest {
...
-              return acc;
-            },
-            {}
-          );
-          // This code adds excludeFromIndexes in the right places
-          addExcludeFromIndexes(entityObject.excludeFromIndexes, entityProto);
-        } else {
-          // This code builds the right entityProto from the entityObject
-          entityProto = entity.entityToEntityProto(entityObject);
-        }
```
danieljbruce, line 1157: This entire code block has been pulled into the `buildEntityProto` function.

src/utils/entity/buildEntityProto.ts
```diff
@@ -0,0 +1,60 @@
...
+import {EntityProtoReduceAccumulator, EntityProtoReduceData} from '../../index';
+import addExcludeFromIndexes = entity.addExcludeFromIndexes;
+
+/**
+ * This function builds the entity proto from the entity object. We cannot
+ * rely on entity.entityToEntityProto for this because this function is only
+ * designed to be used for non-array entities.
+ *
+ */
+export function buildEntityProto(entityObject: Entity) {
```
danieljbruce, line 25: This is the new function that all the code from src/index.ts is being pulled into

test/entity.ts
```diff
@@ -1107,426 +1111,6 @@ describe('entity', () => {
     });
 
     it('should respect excludeFromIndexes', () => {
```
danieljbruce, line 1113: The fixed constants defined below are now going to get reused for the new buildEntityProto function so we pull them out into a separate file so that they can be used here and for `buildEntityProto` tests as well.

test/entity/buildEntityProto.ts
```diff
@@ -0,0 +1,89 @@
...
+// limitations under the License.
+
+import {describe} from 'mocha';
+import {Entities, EntityObject, EntityProto} from '../../src/entity';
+import * as assert from 'assert';
+import {buildEntityProto} from '../../src/utils/entity/buildEntityProto';
+import {
+  entityObject,
+  expectedEntityProto,
+} from '../fixtures/entityObjectAndProto';
```
danieljbruce, line 22: Import the fixed constants that were moved out of `test/entity.ts`.

test/index.ts
```diff
@@ -184,6 +184,12 @@ async.each(
 
       before(() => {
         Object.assign(fakeEntity, fakeEntityInit);
+        const buildEntityProtoModule = proxyquire(
+          '../src/utils/entity/buildEntityProto.js',
+          {
+            '../../entity.js': {entity: fakeEntity},
+          }
+        );
```
danieljbruce, line 192: These tests replace `entity` with `fakeEntity` using `'./entity.js': {entity: fakeEntity},` in the `src/index.ts` file. However, since the code that builds the entity proto used in these tests is no longer in `src/index.ts`, but is in `buildEntityProto.js` instead then the entity in `buildEntityProto.js` needs to be mocked out as well.

daniel-sanche (approved):
LGTM (although I mostly looked at the code, and not the tests)

danieljbruce:
> LGTM (although I mostly looked at the code, and not the tests)

Sounds good. You'll have the opportunity to re-review the tests when all the PRs are put together for a merge onto main.

PR #1361, opened 2024-11-20
fix: Don't throw an error for null inputs to the save function

danieljbruce:
**Summary:**

This PR fixes issue #1327.

**Background:**

Taken directly from [this PR](https://github.com/googleapis/nodejs-datastore/pull/1330)

**Changes:**

`src/entity.ts`: A couple of lines have been added that prevent errors from being thrown when `array` or `entity` is null or undefined. This change may not make the save function perfect and is a sign there may be more work to perfect the save function, but it is clearly an improvement because now code will work for certain inputs that used to throw useless errors.

`test/entity.ts`: A test has been added that throws an error and fails without the new changes to `src/entity.ts`. In this test, `excludeFromIndexes: true` is always in the right place.

**Next steps:**

The save function was [broken down here](https://github.com/googleapis/nodejs-datastore/pull/1356), but for the save function to work properly for every possible input, more edge cases for special cases with nulls, undefined values and other unique values passed into the save function should be explored.


cindy-peng (approved):

kevkim-codes (approved):

danieljbruce:
The checks finally passed for [the copied PR](https://github.com/googleapis/nodejs-datastore/pull/1330) so merging that instead.

PR #1389, opened 2025-05-02
fix: More error visibility in nodejs-datastore

danieljbruce:
Changes in google-gax should be made available in nodejs-datastore so that we can see error stacks and identify root causes for errors going forward.


MarkDuckworth (approved):
LGTM

sofisl (approved):

PR #1390, opened 2025-05-06
test: reproduce commonly cited timeout error

danieljbruce:
## Description

The main intent of this PR is to supply a test confirming that users can see the changes introduced by [this PR](https://github.com/googleapis/gax-nodejs/pull/1740). Primarily, the runQuery test provides an error stack that users can see so that they can identify the root cause of the timeout errors they are seeing. Confirming this behaviour is in place with a test is crucial for solving issues like https://github.com/googleapis/google-cloud-node/issues/7154 going forward.

## Impact

This is a crucial step towards solving https://github.com/googleapis/google-cloud-node/issues/7154 giving us the tools we need to explore client library behaviour when the error from the issue occurs. Having this test in place ensures the impact from the fix in Gax is locked in which will assist users with determining the root cause of errors they are seeing in the future.

## Testing

This PR only adds tests and no source code changes.

`mock-server/datastore-server.ts`: Changes are made to the mock server so that test code can provide its own service configuration providing custom responses for different server grpc endpoints.

`test/mock-server-tester.ts`: This file is meant to be a grpc endpoint agnostic utility file for providing standard grpc services to the mock server. For example, the sendErrorSeries method can be used to make the mock server send a series of errors with the same code and the shutdownServer method shuts down the mock server.

`test/with-runquery-mockserver.ts`: This file runs a test where `runQuery` calls return a deterministic set of `UNAVAILABLE` errors and then it checks the error message to see that it reports the timeout error along with the list of UNAVAILABLE errors in order.

## Additional Information

Some related PRs:
https://github.com/googleapis/gax-nodejs/pull/1740
https://github.com/googleapis/gax-nodejs/pull/1650

## Next Steps

- We have a test for UNAVAILABLE on runQuery now, but we should apply the right fixes for DEADLINE_EXCEEDED errors
- Let's add a test involving transactions


test/try-server.ts
```diff
@@ -17,17 +17,21 @@ import {Datastore} from '../src';
...
+          namespace: `${Date.now()}`,
+          apiEndpoint: 'localhost:50051',
+        });
+        const postKey = datastore.key(['Post', 'post1']);
+        const query = datastore.createQuery('Post').hasAncestor(postKey);
+        const allResults = await datastore.runQuery(query);
+        done();
+      } catch (e) {
+        done(e);
+      }
```
danieljbruce, line 34: The only real change is to make the test runner throw an error if the script throws an error instead of failing silently

test/with-runquery-mockserver.ts
```diff
@@ -0,0 +1,56 @@
...
+            await datastore.runQuery(query);
+            console.log('call failed');
+            assert.fail('The call should not have succeeded');
+          } catch (e) {
+            // The test should produce the right error message here for the user.
+            // TODO: Later on we are going to decide on what the error message should be
+            // The error message is based on client library behavior.
+            assert.strictEqual(
+              (e as Error).message,
+              '4 DEADLINE_EXCEEDED: error details: error count: 1',
```
danieljbruce, line 44: The error message should be something like this. We need to do a bit more investigation into what the client should do in this case to determine the exact error message should be, but for now it's fine to backlog this as a TODO.

test/mock-server-tester.ts
```diff
@@ -0,0 +1,126 @@
+// Copyright 2025 Google LLC
```
danieljbruce, line 1: This file contains utilities for working with the mock server that include a function for producing a deterministic set of errors and a function for shutting down the mock server.

mock-server/datastore-server.ts
```diff
@@ -37,25 +40,38 @@ const packageDefinition = protoLoader.loadSync(PROTO_PATH, {
...
 
-/**
- * Implements the runQuery RPC method.
- */
-function grpcEndpoint(
-  call: {},
-  callback: (arg1: string | null, arg2: {}) => {},
-) {
-  // SET A BREAKPOINT HERE AND EXPLORE `call` TO SEE THE REQUEST.
-  callback(null, {message: 'Hello'});
```
danieljbruce, line 48: This grpcEndpoint function has just been moved inside the only place it is used.

test/mock-server-tester.ts
```diff
@@ -0,0 +1,126 @@
...
+   *
+   * This method creates a `ServiceError` object, simulating an error
+   * that might be returned by a gRPC service. The error includes a
+   * `DEADLINE_EXCEEDED` code (4), a details message indicating the number of
+   * errors generated so far by this instance, and some metadata.
+   *
+   * @returns {ServiceError} A `ServiceError` object representing a simulated
+   *   gRPC error.
+   */
+  generateError() {
```
iam-ethan-huang, line 68: nit - without reading the comment, it's actually hard interpret `generateError` to generate a `DEADLINE_EXCEEDED` error.

test/with-createreadstream-mockserver.ts
```diff
@@ -0,0 +1,51 @@
+// Copyright 2025 Google LLC
```
iam-ethan-huang, line 1: nit - the test file is also a bit confusing. I think your intention is to differentiate the test between non-retryable error like NOT_FOUND v.s. a DEADLINE_EXCEEDED error. 

But the `createreadstream` v.s. `runquery` in the name made it confusing as it seems to do with the functionalities.

danieljbruce: The intent is to differentiate between the calls to the lookup grpc endpoint and the runQuery grpc endpoint.

Note that this PR is in a Draft stage. Although the early feedback is appreciated, I think I'm going to remove this file so that we can lock in the progress we have with the UNAVAILABLE test on runQuery. This gives us the confidence that we can see the error stack for these types of calls.

test/mock-server-tester.ts
danieljbruce: I restructured this so that we pass the error code into generateError. This might make the intent of this method more clear.

mock-server/datastore-server.ts
```diff
@@ -40,22 +44,43 @@ const descriptor = grpc.loadPackageDefinition(packageDefinition);
...
+  ) => void;
 }
 
 /**
  * Starts an RPC server that receives requests for datastore
  */
-export function startServer(cb: () => void) {
+export function startServer<ResponseType>(
+  cb: () => void,
+  serviceConfigurationOverride?: MockServiceConfiguration<ResponseType>,
```
danieljbruce, line 75: This is where a new service can be provided to the mock server.

test/mock-server-tester.ts
```diff
@@ -0,0 +1,91 @@
...
+   * @returns {ServiceError} A `ServiceError` object representing a simulated
+   *   gRPC error.
+   */
+  generateError(code: number) {
+    // SET A BREAKPOINT HERE AND EXPLORE `call` TO SEE THE REQUEST.
+    this.errorSeriesCount++;
+    const metadata = new grpc.Metadata();
+    metadata.set(
+      'grpc-server-stats-bin',
+      Buffer.from([0, 0, 116, 73, 159, 3, 0, 0, 0, 0]),
```
daniel-sanche, line 40: what is this buffer?

test/mock-server-tester.ts
```diff
@@ -0,0 +1,91 @@
...
+    return error;
+  }
+
+  /**
+   * Returns a function that simulates an error response from a gRPC service.
+   *
+   * This method is designed to be used in mock server setups for testing purposes.
+   * It returns a function that, when called, will invoke a callback with a
+   * pre-configured error object, simulating a gRPC service responding with an error.
+   * The error includes a DEADLINE_EXCEEDED code (4), a details message indicating the
```
daniel-sanche, line 55: why DEADLINE_EXCEEDED? Doesn't it take an arbitrary error code?

mock-server/datastore-server.ts
```diff
@@ -12,6 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// eslint-disable-next-line n/no-extraneous-import
+import {ServiceError} from 'google-gax';
+import {Server} from '@grpc/grpc-js';
```
leahecole, line 17: These should probably be added to a package.json instead of disabling extraneous imports

leahecole (commented):
Is there a doc about the mock server tester or the mock server itself? Not sure if this is something your team typically does for your repos, but it seems like a lot of maintenance overhead to introduce. I am seeing yellow flags at reintroducing something somewhat generic that does a series of errors - isn't that what we should be testing in gax? (Which I do know you're working on, I just wonder if this is redundant?)

PR #1394, opened 2025-05-21
test: turn on should import entities

danieljbruce:
## Description

This change makes it so we are not skipping the `should import entities` samples test anymore.

## Impact

Improves our test coverage slightly.

## Testing

This is a test PR. The test used to throw an error saying there was an invalid namespace/kind combination (`Error: 3 INVALID_ARGUMENT: The requested kinds/namespaces are not available`. I suspect this probably has to do with the fact that the kind/namespace needs to be created on the backend, but rather than demand a specific cloud database environment setup for the tests to work I chose to make a slight change in the samples test.

## Additional Information

This test was skipped during the Node 18 upgrade. Sometimes these issues are transient or resolve on their own so we'll see what they do in the CI pipeline.

Note that this changes the sample in GCP documentation.

### DIFF
--- a/samples/import.js
+++ b/samples/import.js
@@ -22,7 +22,7 @@ async function main(file = 'YOUR_FILE_NAME') {
      * TODO(developer): Uncomment these variables before running the sample.
      */
     // const file = 'YOUR_FILE_NAME';
-        
+
     const [importOperation] = await datastore.import({file});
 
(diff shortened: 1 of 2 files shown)


danieljbruce:
We shouldn't change the samples just so we can turn a passing test on.

PR #1412, opened 2025-09-15
test: Ensure datastore transforms are working for transactions

danieljbruce:
## Description

This PR is a follow-up to the PR https://github.com/googleapis/nodejs-datastore/pull/1369 which adds datastore transforms support when the user does datastore.save. It turns out that for saves with transactions no further source code changes are required because transactions call the save method so this PR simply adds some tests for transactions that save with transforms.

## Impact

Ensures datastore transforms work with transactions

## Testing

Added tests where datastore transforms work with transactions.


daniel-sanche (dismissed):
LGTM

daniel-sanche (approved):

